PR #268, opened 2020-08-03
[fix/aut296, fix/aut295, feat/aut311, feat/aut471] - Packaging for macOS (dmg), Linux (deb), Windows (exe), CI for macOS and CI for Windows

ahmedskhalil:
**Description**

Enabling packaging on macOS and Linux platforms, the changes made in [fix/aut295](https://github.com/dmccloskey/SmartPeak2/tree/fix/aut295) were ported into [fix/296](https://github.com/dmccloskey/SmartPeak2/tree/fix/aut296) to ensure the validity of the packaging phase in macOS and Linux in CI.

**Fixes**

- Generating `deb` packages and `dmg` packages without specifying the package type.
- GUI executable is handled differently on each platform (MACOS_BUNDLE on macOS and a regular executable on Windows and Linux).
- Dependencies for `libSmartPeak, libOpenMS, libOpenSwathAlgo` and `libSuperHirn` are not supplied in the `deb` package. Instead, upon installation the user will be prompted to have them fetched and installed by `dpkg`. (Recommended to use `sudo gdebi -i <Package Name>` to resolve any missing dependencies)
- Removed irrelevant `.cmake` files related to packaging.

**Features**

- Packages for macOS (dmg), Linux (deb), Windows (exe) are available after each successful build in Circle CI (Artifacts).

**Future Improvements**

- `build_windows` and `build_macos` pipelines take around ~1h:15m to finish, this is due to OpenMS being built from source (~50min). A speedup can be achieved using caching.

**No longer maintained**

- Target builds for packaging on IDEs is not supported, only on CLI through `cpack`


.circleci/config.yml
```diff
@@ -46,9 +46,73 @@ jobs:
...
             ~/SmartPeak2/.circleci/run_examples.sh ~/SmartPeak2_build/bin
+      - run:
+          command: |
+            set -o xtrace
+            cd ~/SmartPeak2_build
+            cpack -V
+            ls -lh ~/SmartPeak2_build | grep -a .deb
+            dpkg -c ~/SmartPeak2_build/*.deb
+            mkdir -p /tmp/packages
+            cp ~/SmartPeak2_build/*.deb /tmp/packages
```
dmccloskey, line 57: So cool!

.circleci/config.yml
```diff
@@ -46,9 +46,73 @@ jobs:
...
+      - store_artifacts:
+          path: /tmp/packages
+
+  build_macos:
```
dmccloskey, line 61: Nice to see.

CMakeLists.txt
```diff
@@ -240,9 +240,9 @@ set(INSTALL_LIB_DIR ${DEF_INSTALL_LIB_DIR} CACHE PATH "Installation directory fo
 
 # Never use RPATH. We have other means on every platform.
```
dmccloskey, line 241: Please remove the CMAKE lines associated with the Mac workarounds used previously if these are not needed anymore (which I am glad to see they are not).

cmake/package_general.cmake
```diff
@@ -1,86 +1,140 @@
...
+      set(HIDAPI_DIR          "${SDL2_LIB_DIR}/Versions/A/Frameworks")
+	  endif()
+  endif()
 endif (SDL2_FOUND)
+
 find_package(OpenMS REQUIRED)
 if (OpenMS_FOUND)
   ## Add the directory for later calls to CPACK
-  set(OpenMS_LIBRARY_DIR ${OpenMS_DIR}/bin/Release)
+  set(OpenMS_LIBRARY_DIR ${OpenMS_DIR}/lib)
```
dmccloskey, line 82: Can you confirm that this does not break the windows CPack?

src/smartpeak/source/core/Utilities.cpp
```diff
@@ -522,7 +522,7 @@ namespace SmartPeak
       }
 
       const std::string filename(entry.path().filename().string());
-      if (filename == "." || filename == "..") {
+      if (filename == "." || filename == ".." || (*(&filename.at(0))) == '.') {
```
dmccloskey, line 525: Was this the fix for a working executable on mac after calling cpack?

.circleci/config.yml
ahmedskhalil: Thanks for the feedback, I have removed the unused/unrelated files pertaining to the packaging process. It's pass the CI pipelines on both platforms.

For Windows CPack, I will have to perform this somewhere locally while inspecting why AppVeyor is failing with OpenMS build.

CMakeLists.txt
ahmedskhalil: Almost forgotten about this, thanks for reminding.

src/smartpeak/source/core/Utilities.cpp
ahmedskhalil: Yes, to pass the test cases for `Utilities_test` but it didn't seem to have fixed the issue on macOS, I think we need a separate ticket to investigate why `Utilities::getPathnameContent` behaves differently on CircleCi instance.

cmake/package_general.cmake
ahmedskhalil: The `ddl`s inside the bin folder are needed for windows CPack so it has to remain unchanged.

ahmedskhalil:
Packaging is not working for Win CI.

dmccloskey:
> Packaging is not working for Win CI.

Do you mean that packaging on windows in not working in general or is not not automatically being generated on the CI?  If the former, we cannot merge as this is used by our customers in analytics.  If the latter, we can still generate the installer on either my machine or another windows machine from source.

src/smartpeak/include/SmartPeak/core/SharedProcessors.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <cstdlib> // C2039
```
dmccloskey, line 3: Only the include for `string` should be needed.  I don't believe `cstdlib` is needed.

dmccloskey (commented):
Everything still looks good.  If you don't mind addressing my lone comment, resolving the merge conflict, and confirming that CPack still works with windows, then we can merge.  If you have any issues confirming that CPack works on windows let me know and we can discuss how best to get that confirmed and then fixed if need be.

src/smartpeak/include/SmartPeak/core/SharedProcessors.h
ahmedskhalil: Ok, I will try without on the CI

ahmedskhalil:
> > Packaging is not working for Win CI.
> 
> Do you mean that packaging on windows in not working in general or is not not automatically being generated on the CI? If the former, we cannot merge as this is used by our customers in analytics. If the latter, we can still generate the installer on either my machine or another windows machine from source.

The Win CI was working but the Packaging in CircleCI didn't, now the generated packages for all three platforms are present in the artifacts tab in CircleCI.

dmccloskey (approved):
Awesome!

PR #286, opened 2020-11-05
[Feat/aut587] - Migrate all ImGui::columns to ImGui::tables in the GUI

ghost:
**Description**
Migrating all instantiations of the legacy `ImGui::columns` interface to the new `ImGui::tables` interface in the SmartPeak GUI.

**Note**
In `FilePIcker`, calling `ImGui::TableSetupScrollFreeze(column_count, 1);` doesn't seem to have any effect on freezing the headers row : `"Name", "Size", "Type", "Date Modified"`. Freezing the headers row wasn't present prior to this PR.


src/smartpeak/source/ui/FilePicker.cpp
```diff
@@ -64,56 +64,65 @@ namespace SmartPeak
...
-
-    for (int i = 0; static_cast<size_t>(i) < pathname_content_[0].size(); ++i)
+    
+    const int column_count = 4;
+    const char* column_names[column_count] = { "Name", "Size", "Type", "Date Modified" };
+    static ImGuiTableColumnFlags column_flags[column_count] = { ImGuiTableColumnFlags_DefaultSort | ImGuiTableColumnFlags_NoHide, ImGuiTableColumnFlags_NoHide, ImGuiTableColumnFlags_NoHide, ImGuiTableColumnFlags_NoHide
+    };
+    
+    static ImGuiTableFlags table_flags = ImGuiTableFlags_Resizable | ImGuiTableFlags_Hideable | ImGuiTableFlags_Reorderable |
+        ImGuiTableFlags_NoSavedSettings | ImGuiTableFlags_Sortable | ImGuiTableFlags_MultiSortable;
```
dmccloskey: Would you mind adding in `ImGuiTableFlags_NoBordersInBody` to the table flags?  The tables version we are using has a not so nice border in my opinion :)

dmccloskey (changes requested):
Looks great.  I only had an extremely small aesthetic suggestion.  Please re-request a review once ready, and then we can merge.

src/smartpeak/source/ui/FilePicker.cpp
ghost: Thanks for your suggestions, it looks better and simpler without the borders :)

dmccloskey (approved):

PR #297, opened 2021-01-06
[feat/aut177] Allow for resizing the GUI bottom, left and upper panes

bertrandboudaud:
With this modification, it's now possible to resize the left, the bottom and top panels.
I played a little bit with the limitations of imgui. 
The counterpart is some icons appear on the bottom right side of resizable windows, and there are some minor artefacts. Overall, it seems better than before.

To get a better result, the best would to modify imgui to have some flags allowing to resize only some side of the window (for example, left only). Or maybe another way to handle the windowing along with imgui, but it is out of scope for the moment, I think.

image


dmccloskey (approved):
Looks good.  Congratulations on your first PR to SmartPeak!

There was some odd segmentation fault error in a test on MacOS.  I am re-running the pipeline now.  Once all tests pass, I will merge.

PR #298, opened 2021-01-08
Feat/aut700 - Advanced plotting

bertrandboudaud:
- Fixed Sliders
- Disable/Enable legend
- Lines and scatter use the same colors (Compact View)
- More stability to display while running workflow (not fully tested)
- Sliders along with legend checkbox are now part of the chromatogram widget
- Legend is right in order

**compact view**
image

**non compact view**
image

It may be worth noticing that with imGUI when you double right-click on the graph, you have some command items available, like hiding legend.

This code sticks with current architecture (only moved some parts in the widget, which I think is better).
however in general we would need to have the widgets as maintained instances that would avoid these statics, not created when we need to draw (and discard after that). It could be another task.

Another point is it would be very convenient to have a more detailed status of the workflow processing (started, not running, finished ...). For example, I had to play with some flags to know that the graph is ready to display. An event based system or at least a more detailed status would help. Maybe it will come when we will implement the workflow progress.

### DIFF
--- a/src/examples/source/GUI.cpp
+++ b/src/examples/source/GUI.cpp
@@ -101,12 +101,14 @@ int main(int argc, char **argv)
   bool show_chromatogram_legend = true;
   float chromatogram_slider_min = 0.0f;
   float chromatogram_slider_max = 0.0f;
+  bool chromatogram_compact_view = true;
 
   // Spectra display option
   bool spectra_initialized = false;
   bool show_spectra_legend = true;
   float spectra_slider_min = 0.0f;
   float spectra_slider_max = 0.0f;
+  bool spectra_compact_view = true;
 
   // Popup modals
   bool popup_about_ = false;
@@ -967,7 +969,7 @@ int main(int argc, char **argv)
             session_handler_.chrom_x_axis_title, session_handler_.chrom_y_axis_title,
             session_handler_.chrom_time_range.first, session_handler_.chrom_time_range.second, session_handler_.chrom_intensity_min, session_handler_.chrom_intensity_max,
             win_size_and_pos.bottom_and_top_window_x_size_, win_size_and_pos.top_window_y_size_,
-            "Chromatograms Main Window", show_chromatogram_legend, chromatogram_slider_min, chromatogram_slider_max);
+            "Chromatograms Main Window", show_chromatogram_legend, chromatogram_slider_min, chromatogram_slider_max, chromatogram_compact_view);
           plot2d.draw();
           ImGui::EndTabItem();
         }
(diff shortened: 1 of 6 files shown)


src/examples/source/GUI.cpp
```diff
@@ -96,6 +96,20 @@ int main(int argc, char **argv)
   bool show_feature_heatmap_plot = false; // injection vs. feature for a particular metavalue
   bool show_calibrators_line_plot = false; // peak area/height ratio vs. concentration ratio
 
+  // Chromatogram display option
```
dmccloskey, line 99: The GUI options defined at the top of the script are starting to get a bit lengthy.  What are your thoughts on another issues where we look to move these into an e.g., `GUISettings` file?  Or we could add them to the `SessionHandler` object.

src/smartpeak/source/core/SessionHandler.cpp
```diff
@@ -1609,6 +1609,10 @@ namespace SmartPeak
       }
     }
   }
+  void SessionHandler::resetChromatogramRange()
```
dmccloskey, line 1612: Could you add in a unit test for `resetChromatogramRange` and `resetSpectrumRange`?

dmccloskey (changes requested):
As we discussed, it looks quite nice.

If you don't mind adding in the unit tests for the two new methods for the `SessionHandler`, I think it will be good to go.

Also, let me know your thoughts on how best to handle the GUI settings that we define at the top of the `GUI.cpp` file.   I am thinking it is getting close to a time we should look into a more modular solution for those parameters.

src/examples/source/GUI.cpp
bertrandboudaud: Yes I was not sure about where to put that. SessionHandler was one place, i feel uncomfortable putting legend display flag and the others in that object. I'm not sure about the responsibility of the SessionHandler. 
But most of these statics added in this PR, would be better to be non static, and put in the ChromatogramPlotWidget. That would be the right place. instead of creating a ChromatogramPlotWidget on the stack when drawing, we would instanciated it when it is opened.
This would open the way to have multiple chromatograms (in the future, who knows ...).
But it would need some work, little out of scope of this task, so I would recommend to do it in another task, after that.
I would also agree to move it in session handler for the moment, if you feel it's better like this.

src/smartpeak/source/core/SessionHandler.cpp
bertrandboudaud: sure, will do

dmccloskey (approved):

PR #300, opened 2021-01-11
[Feat/aut424] - Advanced Explorer Views 1

ghost:
**Description**

We would like to integrate sorting, searching, and filtering into the tables that are used to display the various views in the SmartPeak GUI.  

**Objectives**
- Ability to sort alphabetically in an ascending and descending order across all tables.
- Showing selected file name in FilePicker.
- Ability to search by name (e.g., sample, sample group, or sequence segment in the Injections view and e.g., transition and transition group in the Transitions view) across all tables (excluding ParametersTable).

**Acceptance criteria**
- Unit tests of internal methods
- Passing CI

**Minor Changes**
- Shortened application name to `SmartPeak`
- Updated `About` section

**Future Improvements**
- Automating build version which can be appended to application name
- Changing display font
- Displaying FilePicker entries in a human readable format


src/smartpeak/source/ui/FilePicker.cpp
```diff
@@ -10,60 +10,99 @@ namespace SmartPeak
 {
   enum ContentColumnID
   {
+    ContentColumnID_ID,
```
bertrandboudaud: What is the use of ID?

ghost: PR created to trigger CI

src/smartpeak/include/SmartPeak/ui/FilePicker.h
```diff
@@ -45,19 +48,23 @@ namespace SmartPeak
...
 
     void run_and_join(
       FilePickerProcessor* processor,
       const std::string& pathname,
       bool& loading_is_done,
       bool& file_was_loaded
     );
+    
+    ///!  rescan pathname_content_ into content_items when needed
+    void update_contents(ImVector<ImDirectoryEntry>& content_items);
```
bertrandboudaud: For code style consistency, shouldn't we rename this updateContents ? (very minor)

src/smartpeak/include/SmartPeak/ui/FilePicker.h
```diff
@@ -45,19 +48,23 @@ namespace SmartPeak
...
 
   public:
     FilePicker()
     {
       // current_pathname_ = fs::current_path().root_path().string();
```
bertrandboudaud: Is it possible to remove these commented lines? or is it interesting to keep it for later?

src/smartpeak/source/core/WorkflowManager.cpp
```diff
@@ -57,6 +57,11 @@ namespace SmartPeak {
   void WorkflowManager::run_and_join(ApplicationHandler& application_handler, bool& done, ApplicationHandler& source_app_handler, const std::set<std::string>& injection_names, const std::set<std::string>& sequence_segment_names, const std::set<std::string>& sample_group_names, const std::vector<ApplicationHandler::Command>& commands)
   {
     // run workflow asynchronously
+    // making local copies
```
bertrandboudaud: I think Douglas already fixed it. I'm not sure it's fixed the same way. may require merging.

src/smartpeak/include/SmartPeak/ui/FilePicker.h
```diff
@@ -45,19 +48,23 @@ namespace SmartPeak
     bool loading_is_done_ = true;
     bool file_was_loaded_ = true;
     bool error_loading_file_ = false;
+    std::atomic_bool files_scanned {false};
```
bertrandboudaud: For code style consistency, should be files_scanned_

src/examples/source/GUI.cpp
```diff
@@ -185,7 +185,7 @@ int main(int argc, char** argv)
   SDL_DisplayMode current;
   SDL_GetCurrentDisplayMode(0, &current);
   SDL_WindowFlags window_flags = (SDL_WindowFlags)(SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI);
-  SDL_Window* window = SDL_CreateWindow("SmartPeak SDL2+OpenGL application", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 1280, 720, window_flags);
+  SDL_Window* window = SDL_CreateWindow("SmartPeak", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 1280, 720, window_flags);
```
dmccloskey: Thanks for updating this

.circleci/config.yml
```diff
@@ -212,13 +212,21 @@ jobs:
...
           msbuild src/SmartPeak_src.sln /verbosity:normal /maxcpucount /p:Configuration=Release
           msbuild src/smartpeak/SmartPeak.sln /verbosity:normal /maxcpucount /p:Configuration=Release
           msbuild src/examples/SmartPeak_class_examples_smartpeak.sln /maxcpucount /property:Configuration=Release
           refreshenv
           & 'C:\Program Files\cmake\bin\ctest.exe' -VV -F
           & 'C:\Program Files\cmake\bin\cpack.exe' -G NSIS64
+      - run:
```
dmccloskey: Great to see this back in.  Can you create a separate PR for this similar to the above or even combined with the above comment?  It would be great to get these integrated ASAP.

.circleci/config.yml
```diff
@@ -95,16 +95,16 @@ jobs:
...
+          cmake -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_STANDARD=14 -DCMAKE_CXX_EXTENSIONS=OFF -DCMAKE_PREFIX_PATH="$(brew --prefix qt5);$(brew --prefix boost)" -DBOOST_USE_STATIC=OFF -DOPENMS_CONTRIB_LIBS=~/OpenMS/contrib_build/ -DHAS_XSERVER=OFF -DWITH_GUI=OFF -DENABLE_TUTORIALS=OFF -DENABLE_DOCS=OFF -DGIT_TRACKING=OFF -DENABLE_UPDATE_CHECK=OFF -DCMAKE_BUILD_TYPE=Release -DPYOPENMS=OFF -DOPENMS_COVERAGE=OFF ..
           make -j4 OpenMS
       - run: |
           cd ~/SmartPeak
           #mkdir superbuild
           mkdir smartpeak_debug_build
           cd superbuild
-          cmake -DUSE_SUPERBUILD=ON -DCMAKE_BUILD_TYPE=Debug .. && make -j4
+          cmake -DUSE_SUPERBUILD=ON -DCMAKE_BUILD_TYPE=Release .. && make -j4
```
dmccloskey: Good to see Release builds here and also for testing.  Can you create a separate PR for this?

src/smartpeak/include/SmartPeak/ui/FilePicker.h
```diff
@@ -45,19 +48,23 @@ namespace SmartPeak
...
+    
+    ///!  rescan pathname_content_ into content_items when needed
+    void update_contents(ImVector<ImDirectoryEntry>& content_items);
 
   public:
     FilePicker()
     {
       // current_pathname_ = fs::current_path().root_path().string();
-      pathname_content_ = Utilities::getPathnameContent(current_pathname_);
+      // pathname_content_ = Utilities::getPathnameContent(current_pathname_);
```
dmccloskey: Is this needed or should it be removed?

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -148,7 +148,7 @@ namespace SmartPeak
       else {
         OpenMS::FileHandler fh;
         LOGI << "Loading: " << filenames.mzML_i;
-        fh.loadExperiment(filenames.mzML_i, chromatograms);
+        fh.loadExperiment(filenames.mzML_i, chromatograms); //TODO:try-catch (SIGABRT)
```
dmccloskey, line 195: What do you mean by this comment?

src/smartpeak/source/core/WorkflowManager.cpp
dmccloskey: I believe these should be removed.  Please check the develop branch.  I would probably recommend merging develop as well to check for any conflicts.

src/smartpeak/source/ui/FilePicker.cpp
```diff
@@ -123,27 +242,45 @@ namespace SmartPeak
...
+          
           const bool is_selected = (selected_entry == row);
           ImGui::TableSetColumnIndex(column);
-          if(ImGui::Selectable(buf, is_selected, selectable_flags))
+          if(ImGui::Selectable(text_buffer, is_selected, selectable_flags))
           {
             selected_entry = row;
-            std::strcpy(selected_filename, pathname_content_[0][selected_entry].c_str());
-            if (ImGui::IsMouseDoubleClicked(0) && pathname_content_[2][selected_entry] == "Directory") // TODO: error prone to rely on strings
+            std::strcpy(selected_filename, Im_directory_entries[selected_entry].Name);
```
dmccloskey: Is this the potential cause of the duplication when selecting the filename in the FilePicker?

src/smartpeak/source/ui/Widget.cpp
```diff
@@ -28,10 +28,134 @@
...
 namespace SmartPeak
 {
+  struct ImTableEntry
+  {
+    int ID;
+    std::vector<const char*> Headers;
+    
+    static const ImGuiTableSortSpecs* s_current_sort_specs;
+    
+    static bool is_digit(const char ch)
```
dmccloskey: Are there any built-in utilities that can be used for this functionality?  I just want to double check that we are not re-inventing the wheel.  If not, then it would be great if we can add some unit testing for these methods.

src/smartpeak/source/ui/Widget.cpp
```diff
@@ -148,7 +243,560 @@ namespace SmartPeak
...
+      }
+      else if (header_name > 0)
+      {
+        cols[header_name] = headers_(header_name - 1).c_str();
+      }
+    }
+    
+    ImGui::Combo("In Column(s)", &selected_col, cols.data(), cols.size() );
+    
+    static ImVector<ImTableEntry> sequence_table_entries;
```
dmccloskey: Is there anyway that the handling of individual `ImVector<ImTableEntry>` and other filter/sorting objects can be refactored into a set of modular methods?  Is it necessary to instantiate all of them here?

dmccloskey (changes requested):
The functionality when I run the application is really nice.  I have a few suggestions on how to make the code a bit cleaner below.  @bertrandboudaud It would be great to get your thoughts as well.

There also appears to be a few bugs introduced here:
1. The declaration of copies during multi-threading execution should be removed as per develop
2. Something appears to be off with the fileninput name.  I am not sure if it is in this branch or in a different branch that is the source of the cause, but I think this should be addressed.

Also, if you don't mind moving the changes to the config.yml to a separate PR so we can get them integrated as soon as possible, that would be great.  I would recommend that also for the changes to the SmartPeak application name and other small fixes that are not a part of the scope for this PR.

.circleci/config.yml
ghost: I added this from develop to avoid merge conflicts

src/smartpeak/include/SmartPeak/ui/FilePicker.h
ghost: They are no longer needed and will be eradicated

src/smartpeak/source/core/RawDataProcessor.cpp
ghost: Occasionally, the `OpenMS::Internal::XMLFile::parse_` fails to parse a given file due to `QDateTime::fromString` being thread-unsafe, it causes an instant crash on macOS and needs to be looked into but is not part of the PR.

src/smartpeak/source/core/WorkflowManager.cpp
ghost: Yes, thanks for reminding

src/smartpeak/source/ui/FilePicker.cpp
ghost: It was because the `picked_pathname_.append()` was called twice in `ImGui::Button("Open")` block

.circleci/config.yml
ghost: Yes I will revert the changes

src/smartpeak/source/ui/Widget.cpp
ghost: None of the implemented functions in `ImTableEntry` are available in the built-in utilities.

src/smartpeak/source/ui/Widget.cpp
ghost: It's necessary to instantiate them here as they get manipulated based on user interaction, sorting/searching cannot be fully modularised as they highly depend on `table_id_` and `ImVector<ImTableEntry>` being updated in the draw method

ghost: Having said that I could encapsulate some of the sub-functionalities in separate methods

bertrandboudaud (commented):
It's very nice, lot of improvements.

I built it but when I run it, I have an exception thrown as soon as I load a sequence (any one). I shared with you, Ahmed. let me know if it's on my side, or if I can test more.

the comments I would add is about the GenericTableWidget::draw. at 2 levels.

1. a lot of repetition, I read you will handle it. the repetition is about the code parts but also there is redundancy about data declaration. Is it possible, at some point to declare only a list of string IDs, and possibly some information attached if needed, that will build all these statics ( ImVector<ImTableEntry>, the *_scanned bolleans), and base the code on the generated map.
The ideal, would be if one day we add a new table we have only one entry to add somewhere, on line vector entry, or one line code.

2. having variable names, specifics strings ids in the GenericTableWidget makes me uncomfortable. Is it possible to externalize these names to the GUI, opening a method on GenericTableWidget, even static, to register all that strings, and ids, from the GUI. I would expect to be able to use a GenericTableWidget (well even if it has a specific behavior that would make sense only in SmartPeak) in another application without touching it: the GUI knows all tables that will be instantiated, they should appear there.

I started to turn some widgets from instantiated-on-draw to a retained-mode (which is not against the immediate mode approach of ImGUI). In this way, you can use variable members. I don't know if it would ease to get rid of these statics if we generalize this (I think we should generalize, but maybe it's not the right moment) however, if it eases the implementation, maybe you can do it like that. For example, passing id in the instantiation of GenericTableWidget.

ghost:
> It's very nice, lot of improvements.
> 
> I built it but when I run it, I have an exception thrown as soon as I load a sequence (any one). I shared with you, Ahmed. let me know if it's on my side, or if I can test more.
> 
> the comments I would add is about the GenericTableWidget::draw. at 2 levels.
> 
> 1. a lot of repetition, I read you will handle it. the repetition is about the code parts but also there is redundancy about data declaration. Is it possible, at some point to declare only a list of string IDs, and possibly some information attached if needed, that will build all these statics ( ImVector, the *_scanned bolleans), and base the code on the generated map.
>    The ideal, would be if one day we add a new table we have only one entry to add somewhere, on line vector entry, or one line code.
> 2. having variable names, specifics strings ids in the GenericTableWidget makes me uncomfortable. Is it possible to externalize these names to the GUI, opening a method on GenericTableWidget, even static, to register all that strings, and ids, from the GUI. I would expect to be able to use a GenericTableWidget (well even if it has a specific behavior that would make sense only in SmartPeak) in another application without touching it: the GUI knows all tables that will be instantiated, they should appear there.
> 
> I started to turn some widgets from instantiated-on-draw to a retained-mode (which is not against the immediate mode approach of ImGUI). In this way, you can use variable members. I don't know if it would ease to get rid of these statics if we generalize this (I think we should generalize, but maybe it's not the right moment) however, if it eases the implementation, maybe you can do it like that. For example, passing id in the instantiation of GenericTableWidget.

Thanks for your feedback, I have tested sequence loading on Windows and it's working fine in all steps. The latest version should have solved these issues.

The repetition is a result of ambiguous behaviour encountered during developing the sorting functionality, loading from non-static and un-scoped variables has resulted text to be displayed from undesignated locations in memory (including displaying OS callstacks), therefore the safest and the most reliable results are achieved when the static variables are declared, instantiated and modified right in the draw function of the respective table. 

Having these variables in a multimap is neat to have but the lookup overhead and the memory consumed by it makes me double it's sustainability in the long run as it's O(log n) for every lookup it triggers in the draw function so this requires further investigation on how to minimise then calls. As for the repetition, major functionalities are now generalised in their own functions. For future roadmap it'd be a good idea to discuss the overall architecture in one of the upcoming tech meetings.

src/examples/source/GUI.cpp
ghost: @dmccloskey I have addressed your requests, please let me know if there's something else that might need improvement.

dmccloskey (changes requested):
Great to see the additional modularity with the new methods.   The repetition still gives me pause.  Maybe that is the only way forward for these features with ImGui.

If you don't mind adding in some unit testing for the new methods, that would be great.  There is quite a bit of complex logic that we would like to quickly check. 

Also, please ensure that @bertrandboudaud takes a look as well.  It would be great to get his opinion.

bertrandboudaud:
> > It's very nice, lot of improvements.
> > I built it but when I run it, I have an exception thrown as soon as I load a sequence (any one). I shared with you, Ahmed. let me know if it's on my side, or if I can test more.
> > the comments I would add is about the GenericTableWidget::draw. at 2 levels.
> > 
> > 1. a lot of repetition, I read you will handle it. the repetition is about the code parts but also there is redundancy about data declaration. Is it possible, at some point to declare only a list of string IDs, and possibly some information attached if needed, that will build all these statics ( ImVector, the *_scanned bolleans), and base the code on the generated map.
> >    The ideal, would be if one day we add a new table we have only one entry to add somewhere, on line vector entry, or one line code.
> > 2. having variable names, specifics strings ids in the GenericTableWidget makes me uncomfortable. Is it possible to externalize these names to the GUI, opening a method on GenericTableWidget, even static, to register all that strings, and ids, from the GUI. I would expect to be able to use a GenericTableWidget (well even if it has a specific behavior that would make sense only in SmartPeak) in another application without touching it: the GUI knows all tables that will be instantiated, they should appear there.
> > 
> > I started to turn some widgets from instantiated-on-draw to a retained-mode (which is not against the immediate mode approach of ImGUI). In this way, you can use variable members. I don't know if it would ease to get rid of these statics if we generalize this (I think we should generalize, but maybe it's not the right moment) however, if it eases the implementation, maybe you can do it like that. For example, passing id in the instantiation of GenericTableWidget.
> 
> Thanks for your feedback, I have tested sequence loading on Windows and it's working fine in all steps. The latest version should have solved these issues.
> 
> The repetition is a result of ambiguous behaviour encountered during developing the sorting functionality, loading from non-static and un-scoped variables has resulted text to be displayed from undesignated locations in memory (including displaying OS callstacks), therefore the safest and the most reliable results are achieved when the static variables are declared, instantiated and modified right in the draw function of the respective table.
> 
> Having these variables in a multimap is neat to have but the lookup overhead and the memory consumed by it makes me double it's sustainability in the long run as it's O(log n) for every lookup it triggers in the draw function so this requires further investigation on how to minimize then calls. As for the repetition, major functionalities are now generalised in their own functions. For future roadmap it'd be a good idea to discuss the overall architecture in one of the upcoming tech meetings.

Please try at least on one table first to have member variables on an instantiated, maintained table as we have seen. I am curious to know why it would not work. I can also have a try (by the way I have tried about FilePicker and the scanned member variable, I had no issue). The chromatogram does like this now (the member variables, for the checkbox and the sliders are actually on the parent class), and also the Parameter table. Once you have tried we can meet and I would like to see why it's not working on mac os. if you don't have time, I can also have a try on my side. once identified, if it's impossible to do this way, we could go with the current implementation.

bertrandboudaud:
I am getting an exception with that branch.

this is the reproduction steps:

1. load the HPLC_UV_Unknowns sequence.
2. run the workflow
3. open the features table
4. open the transition explorer
5. click on the 5-HTP2 transition. The exception will raise in GenericTableWidget::searcher

image

Another call stack, also with the feature table:
1. load the HPLC_UV_Unknowns sequence.
2. run the workflow
3. open the features table
4. open the features explorer
5. click on the "absolute_differrence" checkbox. An exception is raised. but the case stack is differrent.

bertrandboudaud:
> I am getting an exception with that branch.
> 
> this is the reproduction steps:
> 
> 1. load the HPLC_UV_Unknowns sequence.
> 2. run the workflow
> 3. open the features table
> 4. open the transition explorer
> 5. click on the 5-HTP2 transition. The exception will raise in GenericTableWidget::searcher
> 
> image
> 
> Another call stack, also with the feature table:
> 
> 1. load the HPLC_UV_Unknowns sequence.
> 2. run the workflow
> 3. open the features table
> 4. open the features explorer
> 5. click on the "absolute_differrence" checkbox. An exception is raised. but the case stack is differrent.

by the way @ahmedskhalil  sorry, but I see this exception happen on the released version of SmartPeak (dev branch). not sure it needs to be corrected along with this PR.

ghost:
> > I am getting an exception with that branch.
> > this is the reproduction steps:
> > 
> > 1. load the HPLC_UV_Unknowns sequence.
> > 2. run the workflow
> > 3. open the features table
> > 4. open the transition explorer
> > 5. click on the 5-HTP2 transition. The exception will raise in GenericTableWidget::searcher
> > 
> > image
> > Another call stack, also with the feature table:
> > 
> > 1. load the HPLC_UV_Unknowns sequence.
> > 2. run the workflow
> > 3. open the features table
> > 4. open the features explorer
> > 5. click on the "absolute_differrence" checkbox. An exception is raised. but the case stack is differrent.
> 
> by the way @ahmedskhalil sorry, but I see this exception happen on the released version of SmartPeak (dev branch). not sure it needs to be corrected along with this PR.

no worries @bertrandboudaud , it has taken me a while to replicate the exceptions. Regarding the `Widget`'s methods (i.e. `FilterPopup`,  `makeFilters`), they are never used and I am going to remove them or you think we would need them at some point later ?

bertrandboudaud:
> > > I am getting an exception with that branch.
> > > this is the reproduction steps:
> > > 
> > > 1. load the HPLC_UV_Unknowns sequence.
> > > 2. run the workflow
> > > 3. open the features table
> > > 4. open the transition explorer
> > > 5. click on the 5-HTP2 transition. The exception will raise in GenericTableWidget::searcher
> > > 
> > > image
> > > Another call stack, also with the feature table:
> > > 
> > > 1. load the HPLC_UV_Unknowns sequence.
> > > 2. run the workflow
> > > 3. open the features table
> > > 4. open the features explorer
> > > 5. click on the "absolute_differrence" checkbox. An exception is raised. but the case stack is differrent.
> > 
> > 
> > by the way @ahmedskhalil sorry, but I see this exception happen on the released version of SmartPeak (dev branch). not sure it needs to be corrected along with this PR.
> 
> no worries @bertrandboudaud , it has taken me a while to replicate the exceptions. Regarding the `Widget`'s methods (i.e. `FilterPopup`, `makeFilters`), they are never used and I am going to remove them or you think we would need them at some point later ?

I am for removing all dead code. if one day it's needed, it's in the repo 👍

bertrandboudaud:
> > > > I am getting an exception with that branch.
> > > > this is the reproduction steps:
> > > > 
> > > > 1. load the HPLC_UV_Unknowns sequence.
> > > > 2. run the workflow
> > > > 3. open the features table
> > > > 4. open the transition explorer
> > > > 5. click on the 5-HTP2 transition. The exception will raise in GenericTableWidget::searcher
> > > > 
> > > > image
> > > > Another call stack, also with the feature table:
> > > > 
> > > > 1. load the HPLC_UV_Unknowns sequence.
> > > > 2. run the workflow
> > > > 3. open the features table
> > > > 4. open the features explorer
> > > > 5. click on the "absolute_differrence" checkbox. An exception is raised. but the case stack is differrent.
> > > 
> > > 
> > > by the way @ahmedskhalil sorry, but I see this exception happen on the released version of SmartPeak (dev branch). not sure it needs to be corrected along with this PR.
> > 
> > 
> > no worries @bertrandboudaud , it has taken me a while to replicate the exceptions. Regarding the `Widget`'s methods (i.e. `FilterPopup`, `makeFilters`), they are never used and I am going to remove them or you think we would need them at some point later ?
> 
> I am for removing all dead code. if one day it's needed, it's in the repo 👍

Sorry again @ahmedskhalil. This exception is only on this branch. I have been confused, because I had installed the artefact of this branch to confirm the exception, and after installed the artefact of develop branch to compare. but actually when I was testing the develop branch, I was clicking on the GUI.exe  (this branch), not on the SmartPeakGUI.exe. using SmartPeakGUI.exe there is no issue. So this should be fixed, because on windows, it's easy to get this problem. If you have issues reproducing it, I can investigate, let me know.

src/tests/class_tests/smartpeak/source/Widget_test.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+
+BOOST_AUTO_TEST_SUITE(Widget)
+
+BOOST_AUTO_TEST_CASE(widget_constructors)
+{
+  SmartPeak::GenericTableWidget* generictablewidget_ptr = nullptr;
+  SmartPeak::GenericTableWidget* generictablewidget_nullPointer = nullptr;
+  generictablewidget_ptr = new SmartPeak::GenericTableWidget("empty_generic_table");
+  BOOST_CHECK_NE(generictablewidget_ptr, generictablewidget_nullPointer);
```
bertrandboudaud: We can see this pattern to test constructors in many tests, including OpenMS tests, however, new will never return null pointer. it's supposed to throw exceptions in case of error. unless the new handler has been overridden, BOOST_CHECK_NO_THROW would fit better.

src/tests/class_tests/smartpeak/source/Widget_test.cpp
```diff
@@ -0,0 +1,178 @@
...
+  Eigen::Tensor<std::string, 2>     rows_out;
+  
+  getDummyTableEntries(rows_out);
+  SmartPeak::GenericTableWidget TestTable1("TestTable1");
+  TestTable1.updateTableContents(Im_table_entries, is_scanned, rows_out, checkbox_columns);
+  BOOST_CHECK_EQUAL(is_scanned, true);  // updateTableContents is successful
+  
+  // pre-sorting assertion
+  // 1st row
+  BOOST_CHECK_EQUAL(Im_table_entries[0].entry_contents[0], "23dpg.23dpg_1.Heavy");
```
bertrandboudaud, line 121: when accessing entries, it's nice to test if it's available using BOOST_REQUIRE.
For example here
```
BOOST_REQUIRE(Im_table_entries.size>0);  (or the exact size if know it)
BOOST_REQUIRE(Im_table_entries.size[0].entry_contents.size()>0);
```
in this way if the require test fails, the test case will end and the next one will be executed. While without this require statement it will crash. 
It does not make so much difference, but
1) I think without the requirement statement, in case of a crash we will not go to the next test case
2) In release, accessing an out of bound memory can just pass by chance or create a memory corruption, and fail later, which can lead to an unstable test suite.

ghost:
> Great to see the additional modularity with the new methods. The repetition still gives me pause. Maybe that is the only way forward for these features with ImGui.
> 
> If you don't mind adding in some unit testing for the new methods, that would be great. There is quite a bit of complex logic that we would like to quickly check.
> 
> Also, please ensure that @bertrandboudaud takes a look as well. It would be great to get his opinion.

I've added a test file for Widget which will be expanded in part II of the Advanced Explorer Views.

ghost:
Retry DeepCode

dmccloskey (approved):
This will be a much appreciated set of functionalities :).  Thanks for your hard work on this.

PR #302, opened 2021-01-20
[feat/aut680] refactor of handling user parameters

bertrandboudaud:
Not sure to have done things very nice, I have been surprised how parameters impact the code in many place while progressing in the development, and also by the role of the CastValue. At first I wanted to remove this CastValue, but in the end, I kept it and even used it. perhaps it should be renamed "Variant". c++ proposes a variant class but in the C++ 17 version. CastValue handle a lot of casting and parsing (along with utility) which is very difficult to refactor in a short time.

The Parameter panel is a little bit heavy in debug. it could be because of a lot of information displayed, but also because of the refresh rate of the information. It may be improved along with the fix of [AUT-716] as it would deal with the refresh settings.

- Parameter does not use anymore map/vectors/maps, but ParameterSet, container of FunctionParameter, containers of Parameter.

- There are still accessor to underlying map to ease iterator usage and to reduce code modifications.

- It's still possible to initialize with the old structure of map/vector/map.

- Parameter can be initialized from an OpenMS Parameter (to create the schema).

- Parameter can be assigned to another Parameter, the schema, allowing validation.

- Commands return the list of Parameter they need to run (getParameterSchema()), allowing validation when setting up the workflow, not when running it.
*Note*: we could display a message in the Info panel when a parameter is invalid. We can do now with this pull request as soon as the workflow is set. At the moment, to know that a parameter is invalid, the user has to go to the parameter tab.

- Parameter Panel show different colors: user overridden parameters, the default parameters (schema), and the unused (not appearing from the schema)
Parameter2

- Parameter Panel colors invalid values in red. Mouse over it shows some indication: constraint, expected Type
Parameter_3
parameter_1


src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -1129,24 +1163,29 @@ namespace SmartPeak
     LOGD << "END ExtractChromatogramWindows";
   }
 
-  void ExtractSpectraWindows::process(RawDataHandler& rawDataHandler_IO, const std::map<std::string, std::vector<std::map<std::string, std::string>>>& params_I, const Filenames& filenames) const
+  ParameterSet ExtractSpectraWindows::getParameterSchema() const
+  {
+    return ParameterSet();
```
dmccloskey: We could define these and any other "empty" `getParameterSchema` similar to what you had done for `ParameterSet ProcessSequence::getParameterSchema()` for methods where there are no OpenMS parameters.  Please let me know if there are enough "hints" in the Processor class and parameters.csv file to internally define the schema.

I am thinking we could add incorporate this in this PR.  What are your thoughts?

src/smartpeak/source/core/SessionHandler.cpp
```diff
@@ -260,34 +263,71 @@ namespace SmartPeak
...
         LOGD << "Making parameters_table_headers";
-        parameters_table_headers.resize(4);
-        parameters_table_headers.setValues({ "function","name","type","value",/*"default","restrictions","description"*/ });
+        parameters_table_headers.resize(9);
+        parameters_table_headers.setValues({ "function","name","type","value","description","status","valid","restrictions","schema_type" });
       }
       const int n_cols = parameters_table_headers.size();
+
+      // Construct schema based on the current workflow's command list
+      ParameterSet schema_params;
```
dmccloskey, line 299: This is quite nice and will be appreciated by the end user.

src/tests/class_tests/smartpeak/executables.cmake
```diff
@@ -15,6 +15,7 @@ set(core_executables_list
 	SequenceSegmentProcessor_test
 	SessionHandler_test
 	Utilities_test
+	Parameters_test
```
dmccloskey, line 18: Are there any additional tests that are missing for `CastValues` as I think you expanded it quite a bit?

dmccloskey (commented):
This is a great step in the right direction towards a refactoring of how the application is handling user parameters.

My concerns are that 1) we are missing unit test coverage for many of the new methods that were introduced and 2) how we define the default schema for classes where the parameters are defined on the SmartPeak side and not on the OpenMS side.

Please let me know your thoughts on those two points.  I can also help in defining the schemas for point 2 if it is not clear in the code.

src/tests/class_tests/smartpeak/executables.cmake
bertrandboudaud: There is no test for CastValue at all. During the last sprint we planned to do it, but I was about to remove (the day of the sprint, I was trying to do it). In the end no,  and I used it widely instead. so I will reopen the story about testing the CastValue.

bertrandboudaud (commented):
> My concerns are that 1) we are missing unit test coverage for many of the new methods that were introduced 

Agreed. I will do more. For CastValue, it can be another task (another task exists, that I have reopened, AUT-175)

bertrandboudaud:
> 2) how we define the default schema for classes where the parameters are defined on the SmartPeak side and not on the OpenMS side.

There is one example (I found only one case), where we define the default and the minimum for a number of threads.

```  
  ParameterSet ProcessSequence::getParameterSchema()
  {
    std::map<std::string, std::vector<std::map<std::string, std::string>>> param_struct({
    {"SequenceProcessor", {
      {
        {"name", "n_thread"},
        {"type", "int"},
        {"value", "6"},
        {"description", "number of working threads"},
        {"min", "1"}
      }
    }}});
    return ParameterSet(param_struct);
  }
```

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: Ah! yes... yes, schema for non OpenMS should do same as ProcessSequence::getParameterSchema() like you say. I missed this one. Is there a lot? I will make a round.

For empty parameter we can have a more explicit static ParameterSet::empty_set() or use same pattern indeed.

src/tests/class_tests/smartpeak/executables.cmake
dmccloskey: Let's keep the ones for CastValue then as another PR

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: I Have written the parameter schema for 
MergeInjections
mzML
FIAMS
based on the code and the csv files.

However; 2 Parameters appears in the csv files, but I don't find a reference (the use of that) in the code:
FeaturePlotter
SequenceSegmentPlotter

Let me know if I am missing something.

dmccloskey: FeaturePlotter and SequenceSegmentPlotter are legacy.  Please ignore and remove.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -75,7 +75,23 @@ namespace SmartPeak
 
   ParameterSet RawDataProcessor::getParameterSchema() const
   {
-    return ParameterSet();
+    std::map<std::string, std::vector<std::map<std::string, std::string>>> param_struct({
+    {"mzML", {
+      {
+        {"name", "format"},
+        {"type", "string"},
+        {"value", "Chromeleon"},
```
dmccloskey: Most users will be using `XML` so I would recommend using that as the default.

src/smartpeak/include/SmartPeak/core/Parameters.h
```diff
@@ -94,7 +102,17 @@ namespace SmartPeak
 
     Parameter* findParameter(const std::string& parameter);
     const std::string& getFunctionName() const { return function_name_; };
+
+    /**
+    Add a parameter. Doesn't Replace if already exists
```
dmccloskey: Does it make sense to add in a description of the parameters to the docstrings here and below?

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -75,7 +75,23 @@ namespace SmartPeak
...
+        {"description", ""},
+        {"valid_strings", "['Chromeleon','XML']"}
+      },
+      {
+        {"name", "zero_baseline"},
+        {"type", "bool"},
+        {"value", "false"},
+        {"description", ""},
```
dmccloskey: For the description, we can add the following:

"Zeros the baseline of the chromatogram by adjusting all points so that the minimum point is 0."

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -75,7 +75,23 @@ namespace SmartPeak
...
   ParameterSet RawDataProcessor::getParameterSchema() const
   {
-    return ParameterSet();
+    std::map<std::string, std::vector<std::map<std::string, std::string>>> param_struct({
+    {"mzML", {
+      {
+        {"name", "format"},
+        {"type", "string"},
+        {"value", "Chromeleon"},
+        {"description", ""},
```
dmccloskey: We can add the following to the description:

"XML files are those with the ending `.mzML` or `.mzXML`.  Chromeleon files are text files generating from the Thermo family of HPLCs using the Chromeleon software."

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -1641,7 +1703,60 @@ namespace SmartPeak
...
+    {
+      {"name", "min_intensity"},
+      {"type", "float"},
+      {"value", "86000"},
+      {"description", "All features below the minimum intensity will be discarded"},
+    },
+    {
+      {"name", "baseline_type"},
+      {"type", "string"},
+      {"value", "vertical_division"},
```
dmccloskey: The description for these entries are the same as for the parameters of the PeakPickerMRM.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -1994,9 +2109,30 @@ namespace SmartPeak
...
+        {"type", "string"},
+        {"value", "norm_sum"},
+        {"description", ""},
+        {"valid_strings", "['norm_sum','norm_max']"}
+      },
+      {
+        {"name", "feature_name"},
+        {"type", "string"},
+        {"value", "intensity"},
+        {"description", ""},
```
dmccloskey: We can use the following for the description:

"The name of the FeatureMap attribute to use.  Examples include `peak_apex_int`, `peak_area`, and `intensity`."

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2048,9 +2184,24 @@ namespace SmartPeak
...
 
+  ParameterSet IsotopicCorrections::getParameterSchema() const
+  {
+    std::map<std::string, std::vector<std::map<std::string, std::string>>> param_struct({
+    {"IsotopicCorrections", {
+      {
+        {"name", "correction_matrix_agent"},
+        {"type", "string"},
+        {"value", "TBDMS"},
+        {"description", ""},
```
dmccloskey: "The correction matrix corresponding to the derivatization agent used when processing the samples for LC-MS/MS or GC-MS."

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2093,9 +2244,27 @@ namespace SmartPeak
...
   }
 
+  ParameterSet CalculateIsotopicPurities::getParameterSchema() const
+  {
+    std::map<std::string, std::vector<std::map<std::string, std::string>>> param_struct({
+    {"CalculateIsotopicPurities", {
+      {
+        {"name", "isotopic_purity_values"},
+        {"type", "string"},
+        {"value", ""},
```
dmccloskey, line 2244: Please ask @ahmedskhalil for a description of these parameters.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2170,9 +2339,22 @@ namespace SmartPeak
...
   }
 
+  ParameterSet CalculateMDVAccuracies::getParameterSchema() const
+  {
+    std::map<std::string, std::vector<std::map<std::string, std::string>>> param_struct({
+    {"CalculateMDVAccuracies", {
+      {
+        {"name", "feature_name"},
+        {"type", "string"},
+        {"value", ""},
```
dmccloskey, line 2341: The same description as the previous can be used here.

src/smartpeak/source/core/SampleGroupProcessor.cpp
```diff
@@ -32,7 +32,55 @@ namespace SmartPeak
...
   ParameterSet MergeInjections::getParameterSchema() const
   {
-    return ParameterSet();
+    std::map<std::string, std::vector<std::map<std::string, std::string>>> param_struct({
+    {"MergeInjections", {
+      {
+        {"name", "scan_polarity_merge_rule"},
+        {"type", "string"},
+        {"value", "WeightedMean"},
+        {"description", ""},
```
dmccloskey: "The method used to merge the different feature values.  Options include `Sum` (the sum of all feature values), `Min` (the smallest feature value), `Max` (the largest feature value), `Mean` (the mean of all feature values), and `WeightedMean` (the weight average of all feature values)."

src/smartpeak/source/core/SampleGroupProcessor.cpp
```diff
@@ -32,7 +32,55 @@ namespace SmartPeak
...
+      {
+        {"name", "scan_polarity_merge_feature_name"},
+        {"type", "string"},
+        {"value", "peak_apex_int"},
+        {"description", ""}
+      },
+      {
+        {"name", "mass_range_merge_feature_name"},
+        {"type", "string"},
+        {"value", "peak_apex_int"},
```
dmccloskey, line 67: The same description for the `feature_name` can be used here as well.

dmccloskey (changes requested):
Looks very nice.  I added a few small comments that should be addressed at your discretion as well as text for the descriptions.

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: actually i feel these 2 baseline_type and integration_type should not be listed here as they are not used later in the code, and they are defined in openms PeakIntegrator, which is included in the PickMS1Features::getParameterSchema

dmccloskey: If that is the case then please just remove them.

dmccloskey (approved):
Great

PR #303, opened 2021-01-21
[feat/aut716] Fixed Chromatogram stops updating after having selected…

bertrandboudaud:
To have a proper refresh and fix the issue we need to have the status of what has changed:

- The model has changed?
- The input has changed?

regarding the input, instead of keeping this information in the SessionHandler, I would put this information in the chromatogram view: what has been used to extra the data from the model (what the user has selected). Same for the time range, I would put also it in the view, instead of keeping it in the SessionHandler. If one of these parameter change, the view update getting the data from the model.

To know if the model has changed, at the moment it still uses the working workflow status GUI.cpp. This is not so nice, but the code involved is not huge or problematic.

getChromatogramScatterPlot returns a structure, ChromatogramScatterPlot,  the result is stored in the view to avoid updates. The structure is generic enough to be used by the spectrum view (which is the same class as chromatogram) - the spectrum view is not tested yet in this PR.

in GUI.cpp, the chromatogramview is created and maintained instead of being instantiated in each frame. in this way it can hold information that is tight to the view in the instance: the input used to do the graph, the slider values, and max ...

I wanted to have your feedback on this because I'm not sure about SessionHandler, what is the role. For me, it's the model in the MVC design pattern. But then, I'm not sure chromatogram data should be in the model or it's better if we can, to generate on-demand with input parameter that fit the user, like in this PR.

This PR also open the way to have multiple chromatogram views more easily.

If this is ok for you, I suggest doing this for chromatogram and spectrum only, maybe plan such changes for another view when modification will be needed.

### DIFF
--- a/src/examples/source/GUI.cpp
+++ b/src/examples/source/GUI.cpp
@@ -10,6 +10,10 @@
 #include <SmartPeak/core/SessionHandler.h>
 #include <SmartPeak/ui/FilePicker.h>
 #include <SmartPeak/ui/GuiAppender.h>
+#include <SmartPeak/ui/Heatmap2DWidget.h>
+#include <SmartPeak/ui/CalibratorsPlotWidget.h>
+#include <SmartPeak/ui/ChromatogramPlotWidget.h>
+#include <SmartPeak/ui/SpectraPlotWidget.h>
 #include <SmartPeak/ui/Report.h>
 #include <SmartPeak/ui/Workflow.h>
 #include <SmartPeak/ui/WindowSizesAndPositions.h>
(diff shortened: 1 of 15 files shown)


dmccloskey (commented):
I think conceptual this is a step in the right direction.  As you say, I can see clearly how this would allow for the user to save or filter through multiple views simultaneous, which would be helpful during the review process.

I also like the encapsulation of all elements into the widget itself.  I would recommend making a separate file file each widget from now on if we are going to use this as a template for the current widgets and future widgets.

My only concern is the logic used for deciding the refresh and how it interacts with the user filtering.  However, I think if you could walk me through the code, that would be helpful to my understanding.

bertrandboudaud:
- refresh is not perfect, maybe wait for an event based system. story will be created.
- we didn't talked about the proposition to have one file per widget, I like I will do for this pull request.
- agreed about the removal of the data that we can remove as it has been done here from SessionHandler. and SessionHandler will converge to a data model and a data filters provider.

src/examples/source/GUI.cpp
```diff
@@ -10,6 +10,10 @@
 #include <SmartPeak/core/SessionHandler.h>
 #include <SmartPeak/ui/FilePicker.h>
 #include <SmartPeak/ui/GuiAppender.h>
+#include <SmartPeak/ui/Heatmap2DWidget.h>
```
dmccloskey, line 13: Nice.

dmccloskey (approved):
Looks good from my end.  Once the CI passes we can merge.

PR #304, opened 2021-01-25
[feat/aut169] Add getID, getName and getDescription to SequenceProcessor

ChristopherAbram:
✅ Add `IProcessorDescription` interface,
✅ Implement `IProcessorDescription` interface in `SequenceProcessor` classes,
✅ Add unit tests for relevant `SequenceProcessor`'s getters.

📌 Disclaimer:
- Assigned IDs are now set to `-1`, this probably needs an update,
- Descriptions might be not descriptive enough - I copied class docs to the string,
- Since I created `IProcessorDescription` interface, I can also update other Processor classes with this interface (out of the scope of this ticket)
- Maybe location of `IProcessorDescription` can be under new directory `interface` or `iface`:
image

### DIFF
--- a/src/smartpeak/include/SmartPeak/core/SequenceProcessor.h
+++ b/src/smartpeak/include/SmartPeak/core/SequenceProcessor.h
@@ -28,6 +28,7 @@
 #include <SmartPeak/core/SequenceHandler.h>
 #include <SmartPeak/core/SequenceSegmentProcessor.h>
 #include <SmartPeak/core/SampleGroupProcessor.h>
+#include <SmartPeak/iface/IProcessorDescription.h>
 #include <map>
 #include <memory> // shared_ptr
 #include <set>
@@ -99,7 +100,7 @@ namespace SmartPeak
     const std::vector<std::shared_ptr<RawDataProcessor>>& methods
   );
 
-  struct SequenceProcessor {
+  struct SequenceProcessor : IProcessorDescription {
     SequenceProcessor(SequenceHandler& sh) : sequenceHandler_IO(&sh) {}
     virtual ~SequenceProcessor() = default;
 
@@ -119,6 +120,11 @@ namespace SmartPeak
     CreateSequence() = default;
     CreateSequence(SequenceHandler& sh) : SequenceProcessor(sh) {}
     void process() const override;
+
+    /* IProcessorDescription */
+    int getID() const override { return -1; }
+    std::string getName() const override { return "CREATE_SEQUENCE"; }
+    std::string getDescription() const override { return "Create a new sequence from file or wizard"; }
   };
 
   /**
@@ -133,6 +139,11 @@ namespace SmartPeak
     ProcessSequence(SequenceHandler& sh) : SequenceProcessor(sh) {}
     static ParameterSet getParameterSchema();
     void process() const override;
+
+    /* IProcessorDescription */
+    int getID() const override { return -1; }
+    std::string getName() const override { return "PROCESS_SEQUENCE"; }
+    std::string getDescription() const override { return "Apply a processing workflow to all injections in a sequence"; }
   };
 
   /**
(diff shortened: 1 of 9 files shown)


bertrandboudaud:
Yes, would be nice to use this interface for other processors. I'm not sure it's so out of scope of the ticket. since you introduce an interface, it may impact other class. is it lot of work? i feel if it's very easy you could go ahead. if not create a story once it is merged.

about the interface directory, yes iface is nice. but if we do so, all the interfaces should go there. would be better to put them all at the same time. perhaps it would be better to stay like you have done and create a story about that ("move all interfaces in dedicated directory").

src/smartpeak/include/SmartPeak/core/SequenceProcessor.h
```diff
@@ -124,6 +135,11 @@ namespace SmartPeak
     ProcessSequenceSegments() = default;
     ProcessSequenceSegments(SequenceHandler& sh) : SequenceProcessor(sh) {}
     void process() const override;
+
+    /* IProcessorDescription */
+    int getID() const { return -1; }
```
bertrandboudaud: "override" would be welcome for all these getID() getName() and getDescription()

src/smartpeak/include/SmartPeak/core/IProcessorDescription.h
```diff
@@ -0,0 +1,27 @@
...
+
+        /**
+            Get the processor struct description
+        */
+        virtual std::string getDescription() const = 0;
+
+        virtual ~IProcessorDescription() = default;
+    };
+
+    typedef std::shared_ptr<IProcessorDescription> IProcessorDescriptionPtr;
```
bertrandboudaud: This declaration can be usefull, but I would prefer a team agreement and to have a check on all code to use this pattern in a dedicated story to have consistent code. my concern is to have this pattern use only on some parts of code and not on others. to me explicit shared_ptr is nice because you can also have unique_ptr the declaration my hide this.

bertrandboudaud (changes requested):

src/smartpeak/include/SmartPeak/core/SequenceProcessor.h
ChristopherAbram: Good point, `override` well describe developer's intention, however its optional

src/smartpeak/include/SmartPeak/core/IProcessorDescription.h
ChristopherAbram: Agree with your concerns, its a good practice to be consistent and keep same `coding conventions` everywhere in the project. We can have some discussion later during sprint review/retro.

ChristopherAbram:
Should not be much work to add `IProcessorDescription` interface to another `Processor` classes. 
I am not sure however what you mean by moving all interface to one directory - at least if it comes to processor classes, I can see plenty of abstract classes, but not many classes that have only pure virtual functions. Moreover, these classes have very close `process()` function signatures, but still different. I think such work of defining interfaces for these classes would be, as you mentioned, another task, refactoring of code base

bertrandboudaud:
I haven't check if there are other interfaces - sorry. If there is no other interfaces, you can go create the iface directory and put yours inside. I just would like (it's my point of view) not to have some interfaces in one directory, others (if they exist) in another.

bertrandboudaud (approved):

src/smartpeak/include/SmartPeak/iface/IProcessorDescription.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include <memory>
+
+namespace SmartPeak 
+{
+    struct IProcessorDescription
```
dmccloskey: Can you change to two spaces instead of tabs?

src/smartpeak/include/SmartPeak/iface/IProcessorDescription.h
```diff
@@ -0,0 +1,27 @@
...
+
+        /**
+            Get the processor struct description
+        */
+        virtual std::string getDescription() const = 0;
+
+        virtual ~IProcessorDescription() = default;
+    };
+
+    typedef std::shared_ptr<IProcessorDescription> IProcessorDescriptionPtr;
```
dmccloskey: Is this typdef used anywhere or can it be removed?

dmccloskey (changes requested):
Interesting choice to implement an interface pattern for the processor classes.

If you don't mind taking care of my small comments I think we will be good to go once the CI is working.

dmccloskey (approved):

PR #305, opened 2021-01-27
Add test for ApplicationProcessor classes

ChristopherAbram:
The unit test for `ApplicationProcessor` classes:
- I observed that `FilePickerProcessor` classes succeed when a path associated with an object is empty or point to wrong file
   - Success when path is empty -> should fail (return `false`),
   - Success when path points to non-existing file -> should fail (return `false`),
   - Some classes throws an exception when path points to file which exists but have wrong structure -> perhaps the exception should be caught in `process()` and the method should return false
- The `ApplicationProcessors::processCommands()` is not tested. The only idea I have is to verify exception handling. The thing is that this procedure relies on `SequenceProcessor` classes, so this test would mostly test dependencies, not the actual procedure.

📌Please give me hints about `FilePickerProcessor`, at the moment implementation requires some fixes.
❗ Because we don't use mocking approach, I created a test fixture which creates default objects (`default-constructible`) as stubs. However, this way breaks rules of unit testing with regards to isolating dependencies, which in turn makes them class-level integration tests.
❗ Many test fails at this stage.

### DIFF
--- a/src/tests/class_tests/smartpeak/source/ApplicationProcessor_test.cpp
+++ b/src/tests/class_tests/smartpeak/source/ApplicationProcessor_test.cpp
@@ -25,15 +25,108 @@
 
 #define BOOST_TEST_MODULE ApplicationProcessor test suite
 #include <boost/test/included/unit_test.hpp>
+#include <boost/filesystem.hpp>
 #include <SmartPeak/core/ApplicationProcessor.h>
+#include <SmartPeak/core/SequenceProcessor.h>
+#include <SmartPeak/core/Filenames.h>
 #include <boost/filesystem.hpp>
 
 using namespace SmartPeak;
 using namespace std;
+namespace fs = boost::filesystem;
 
-BOOST_AUTO_TEST_SUITE(applicationprocessor)
+struct ApplicationProcessorFixture 
+{
+  /* ctor/dtor */
+  ApplicationProcessorFixture() 
+  {
+    datapath_ = SMARTPEAK_GET_TEST_DATA_PATH("");
+    auto workflow = fs::path{ datapath_ } / fs::path{ "workflow_csv_files" };
 
-BOOST_AUTO_TEST_CASE(buildcommandsfromids)
+    filenames_ = Filenames::getDefaultStaticFilenames(workflow.string());
+    filenames_.referenceData_csv_i = (fs::path{ datapath_ } / fs::path{ "MRMFeatureValidator_referenceData_1.csv" }).string();
+
+    // Injections sequence:
+    std::vector<InjectionHandler> seq;
+    {
+      seq.push_back(InjectionHandler{});
+      seq.push_back(InjectionHandler{});
+    }
+    ah_.sequenceHandler_.setSequence(seq);
+    // Sequence segments:
+    std::vector<SequenceSegmentHandler> seq_seg;
+    {
+      seq_seg.push_back(SequenceSegmentHandler{});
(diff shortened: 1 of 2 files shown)


src/tests/class_tests/smartpeak/source/ApplicationProcessor_test.cpp
```diff
@@ -4,14 +4,112 @@
...
+#include <SmartPeak/core/Filenames.h>
 
 using namespace SmartPeak;
 using namespace std;
+namespace fs = boost::filesystem;
 
-BOOST_AUTO_TEST_SUITE(applicationprocessor)
+struct ApplicationProcessorFixture 
+{
+	/* ctor/dtor */
```
bertrandboudaud: You used tabs everywhere, it seems the other files are using 2 spaces indentation.

bertrandboudaud:
For FilePickerProcessor I agree that for a non-existing file (or empty), it should return false.
However; FilePickerProcessor may be used to write a file (?) in that case maybe not...

About the Exception ... this is not easy, don't know if it has already been discussed.. for me, an exception should be caught only at some point if the catcher can handle it: he knows what is the exception, how critical it is. a general rule (catch everything in process()). I am not sure. It's probably nice to catch most of the time, but may be dedicated catch would be nice. So the rule would be : if the Processor can catch the error and treat it without putting the data model into a bad state, yes, catch and return false (again, it should be most of the case). else throw.

src/tests/class_tests/smartpeak/source/ApplicationProcessor_test.cpp
```diff
@@ -4,14 +4,112 @@
...
+public:
+	ApplicationHandler ah_;
+	std::string datapath_;
+	Filenames filenames_;
+};
+
+BOOST_FIXTURE_TEST_SUITE(ApplicationProcessor, ApplicationProcessorFixture)
+
+/* processCommands */
+//BOOST_AUTO_TEST_CASE(buildcommandsfromids)
```
dmccloskey: Is there a reason for the commented code here?

dmccloskey (changes requested):
Great work getting this tests in place!  The test fixture is a fine way to reduce redundant loading of the class.  The use of other dependencies to load in the data, etc. is fine from my point of view given the complexity (or lack thereof) of this software application so long as there is test coverage across all dependencies in such a way that the fault can be quickly isolated if something goes wrong.

I agree with Bertrand in regards to the questions about the `FilePickerProcessor` class.

Please be sure to change the spacing to two spaces instead of tabs.

Please add in the unit test for ApplicationProcessors::processCommands() but you can omit a test for a value.  We can simply check to make sure it runs without throwing an error.

src/tests/class_tests/smartpeak/source/ApplicationProcessor_test.cpp
ChristopherAbram: I guess I should have changed default settings in Visual Studio 2019 ¯\\_(ツ)_/¯

src/tests/class_tests/smartpeak/source/ApplicationProcessor_test.cpp
ChristopherAbram: No need for it, this is still WIP

ghost (commented):
Looking good from my side once CI for Ubuntu are passed.

dmccloskey (approved):

PR #307, opened 2021-02-02
Feat/aut425

bertrandboudaud:
Ability to save and load workflows

The Load and Save are accessible from the File Menu (File->Export, File->Import), and also the workflow is loaded along with the sequence if workflow.csv is present.

The workflow window now uses the modal dialog content. As a consequence

- There is no two views anymore, one to edit, one to view, but only one.
- removed the edit workflow from the edit menu (the edit menu is useful?)

This is out of scope of the task, but it was easy and it has also simplified the development. If you feel it's too much out of scope, I can revert this.

2021-02-02 14_55_27-WorkflowWindow
 
Note! I have included the temporary(?) patch from @ahmedskhalil  about thread crashing because it crahed on my side, and I add some code around the patch. so I need this patch at the moment.


src/smartpeak/include/SmartPeak/core/SequenceHandler.h
```diff
@@ -53,6 +53,10 @@ namespace SmartPeak
       const OpenMS::FeatureMap& featureMap_I
     );
 
+    void setWorkflow(std::vector<std::string>& commands);
```
dmccloskey: Should the input parmeter be constant qualified?

src/examples/data/FIAMS_FullScan_Unknowns/workflow.csv
```diff
@@ -0,0 +1,14 @@
+command_name
```
dmccloskey, line 1: Nice.

src/examples/source/GUI.cpp
```diff
@@ -316,22 +316,37 @@ int main(int argc, char **argv)
         for (const std::string& pathname : {application_handler_.mzML_dir_, application_handler_.features_in_dir_, application_handler_.features_out_dir_}) {
           fs::create_directories(fs::path(pathname));
         }
-        for (ApplicationHandler::Command& cmd : application_handler_.commands_)
-        {
-          for (std::pair<const std::string, Filenames>& p : cmd.dynamic_filenames)
+        std::vector<ApplicationHandler::Command> commands;
+        bool commands_created = true;
+        CreateCommand createCommand(application_handler_);
+        for (const auto& command_name : application_handler_.sequenceHandler_.getWorkflow()) {
```
dmccloskey: Nice.  I am thinking we made a good decision in moving towards a string representation of the workflow steps that are converted on the fly as shown here.  I think this will make "headless" mode much easier to implement.

On that note, do you think it would make sense to refactor this chunk of code into its own method?

src/examples/source/GUI.cpp
```diff
@@ -409,6 +420,12 @@ int main(int argc, char **argv)
             popup_file_picker_ = true;
             update_session_cache_ = true;
           }
+          if (ImGui::MenuItem("Workflow")) {
```
dmccloskey, line 415: Nice

src/smartpeak/source/core/WorkflowManager.cpp
```diff
@@ -47,15 +47,21 @@ namespace SmartPeak {
     return done_;
   }
 
-  void WorkflowManager::run_and_join(ApplicationHandler& application_handler, bool& done, ApplicationHandler& source_app_handler, const std::set<std::string>& injection_names, const std::set<std::string>& sequence_segment_names, const std::set<std::string>& sample_group_names)
+  void WorkflowManager::run_and_join(ApplicationHandler& application_handler, bool& done, ApplicationHandler& source_app_handler, const std::set<std::string>& injection_names, const std::set<std::string>& sequence_segment_names, const std::set<std::string>& sample_group_names, const std::vector<ApplicationHandler::Command>& commands)
   {
+    // making local copies
```
dmccloskey, line 52: Good fix.

src/smartpeak/source/ui/Workflow.cpp
```diff
@@ -64,30 +56,101 @@ namespace SmartPeak
       {
         if (ImGui::Selectable(s))
         {
-          BuildCommandsFromNames buildCommandsFromIds(*application_handler_);
-          std::string ids;
+          std::vector<std::string> ids;
           const std::string s_string { s };
           if (s_string == "LCMS MRM Unknowns")
-            ids = "LOAD_RAW_DATA MAP_CHROMATOGRAMS PICK_MRM_FEATURES QUANTIFY_FEATURES CHECK_FEATURES SELECT_FEATURES STORE_FEATURES";
+            ids = { "LOAD_RAW_DATA",
```
dmccloskey, line 62: This looks much nicer

src/smartpeak/source/ui/Workflow.cpp
```diff
@@ -96,76 +159,63 @@ namespace SmartPeak
...
       for (const auto& p : n_to_raw_data_method_)
       {
         if (ImGui::Selectable(p.second->getName().c_str()))
         {
-          createCommand.name_ = p.second->getName();
-          const bool created = createCommand.process();
-          if (created) {
-            commands_.push_back(createCommand.cmd_);
-          }
+          application_handler_->sequenceHandler_.getWorkflow().push_back(p.second->getName());
```
dmccloskey, line 166: This consolidated quite a bit of code.

src/tests/class_tests/smartpeak/source/ApplicationProcessor_test.cpp
```diff
@@ -246,4 +247,59 @@ BOOST_AUTO_TEST_CASE(SetOutputFeaturesPathnameGetName)
...
+    "PICK_MRM_FEATURES",
+    "QUANTIFY_FEATURES",
+    "CHECK_FEATURES",
+    "SELECT_FEATURES",
+    "STORE_FEATURES"
+  };
+  application_handler.sequenceHandler_.setWorkflow(command_names);
+  StoreSequenceWorkflow application_processor(application_handler);
+  BOOST_CHECK_EQUAL(application_processor.getName(), "StoreSequenceWorkflow");
+  application_processor.pathname_ = (fs::temp_directory_path().append(fs::unique_path().string())).string();
```
dmccloskey, line 264: Is boost filesystem needed here or can the test be accomplished using only the `SMARTPEAK_GET_TEST_DATA_PATH` macro?  This is quite minor so please address at your discretion.

This comment also applies to the similar test below.

dmccloskey (commented):
This looks really nice!  That was a very good idea to combine the modal with the table view.  We should definitely keep the loading/storing of the workflow steps in this PR even though it is out of scope (although quite related).  I can't wait to start using it.

I made some very minor comments.  Please let me know your thoughts.

bertrandboudaud:
The requested changes have been applied in the last commit.

src/smartpeak/source/core/ApplicationProcessor.cpp
```diff
@@ -655,19 +655,18 @@ namespace SmartPeak
...
-      if (created) {
+    CreateCommand createCommand(application_handler_);
+    for (const auto& command_name : names_) {
+      createCommand.name_ = command_name;
+      bool commands_created = createCommand.process();
+      success &= commands_created;
+      if (commands_created) {
         commands_.push_back(createCommand.cmd_);
-      }
+      } // if not, no need to log. createCommand already logs an error.
```
dmccloskey, line 667: I didn't catch this the first round.  Good catch!

src/smartpeak/source/core/ApplicationProcessor.cpp
```diff
@@ -655,19 +655,18 @@ namespace SmartPeak
...
     }
-    return true;
+    return success;
```
dmccloskey, line 669: I missed this as well.  Good Catch.

dmccloskey (approved):
Great work.

PR #312, opened 2021-02-06
[fix/aut809] - File name not shown when selected

ghost:
**Features**

- Displaying file name when selected in the designated field in the FilePicker popup-window.
- Adding double-click-to-open feature to open files either by clicking on "open" button or double clicking on the file.

**Testing**
- Successful manual testing of the added features/fixes in the GUI.

### DIFF
--- a/src/smartpeak/source/ui/FilePicker.cpp
+++ b/src/smartpeak/source/ui/FilePicker.cpp
@@ -38,6 +38,7 @@ namespace SmartPeak
     }
 
     static int selected_entry = -1;
+    static char selected_filename[256] = {0};
 
     if (ImGui::Button("Up"))
     {
@@ -46,6 +47,7 @@ namespace SmartPeak
         current_pathname_ = parent;
       }
       pathname_content_ = Utilities::getPathnameContent(current_pathname_);
+      memset(selected_filename, 0, sizeof selected_filename);
       selected_entry = -1;
     }
     ImGui::SameLine();
@@ -65,6 +67,7 @@ namespace SmartPeak
       {
         current_pathname_.assign(new_pathname);
         pathname_content_ = Utilities::getPathnameContent(current_pathname_);
+        memset(selected_filename, 0, sizeof selected_filename);
         selected_entry = -1;
         ImGui::CloseCurrentPopup();
       }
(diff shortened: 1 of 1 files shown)


src/smartpeak/source/ui/FilePicker.cpp
```diff
@@ -128,19 +130,37 @@ namespace SmartPeak
...
+              {
+                if (picked_pathname_.back() != '/')
+                {
+                  picked_pathname_.append("/");
+                }
+                picked_pathname_.append(pathname_content_[0][selected_entry]);
+              }
+              filter.Clear();
+              selected_entry = -1;
+              LOGI << "Picked pathname: " << picked_pathname_;
```
bertrandboudaud, line 163: It's missing
```
runProcessor();
clearProcessor();
```
so that when you double click, you will execute the processor

src/smartpeak/source/ui/FilePicker.cpp
```diff
@@ -151,7 +171,10 @@ namespace SmartPeak
     ImGui::Separator();
 
     ImGui::PushItemWidth(ImGui::GetContentRegionAvail().x * 0.9f);
-    ImGui::InputTextWithHint("", "File name", selected_filename, IM_ARRAYSIZE(selected_filename));
+    if (ImGui::IsMouseClicked(0))
+      ImGui::InputTextWithHint("", "File name", selected_filename, IM_ARRAYSIZE(selected_filename));
```
bertrandboudaud: The line is exactly the same as the other one below ?

bertrandboudaud (changes requested):
Would be nice to be able to use the Filepicker to save files. that is, to "select" a file that does not exists: using the File Name field.
it's possible by replacing 

```
      if (selected_entry >= 0)
      {
        if (picked_pathname_.back() != '/') // do not insert "/" if current_pathname_ == root dir, i.e. avoid "//home"
        {
          picked_pathname_.append("/");
        }
        picked_pathname_.append(pathname_content_[0][selected_entry]);
      }
```

with
```
//      if (selected_entry >= 0)
      {
        if (picked_pathname_.back() != '/') // do not insert "/" if current_pathname_ == root dir, i.e. avoid "//home"
        {
          picked_pathname_.append("/");
        }
        picked_pathname_.append(selected_filename);
      }
```
That is to use the Filename field instead of getting the filename from the selected item. is it possible to do in this PR?

bertrandboudaud (approved):

PR #315, opened 2021-02-09
[feat/aut358] Peak picking for MS2 RawDataProcessor method

bertrandboudaud:
Created a new RawDataProcessor struct called PickMS2Features

Created a new processor method that wraps a series of method calls defined in
https://github.com/OpenMS/OpenMS/blob/develop/src/topp/FeatureFinderMetabo.cpp where the methods for MassTraceDetection().run, ElutionPeakDetection().detectPeaks, and then ff = FeatureFindingMetabo(); ff.run are called in order.  Note that the output will be the input for RawDataProcessor::SearchAccurateMasss.  Please refer to the python code from Efi as well.

Provided converters from the SmartPeak Parameters structure to the OpenMS Parameters structure similar to the other RawDataProcessor methods so that the user can configure the OpenMS classes prior to calling.

Included logging and crash prevention similar to others.

The current workflow to reproduce Efi's python script is:
image

There is a fixed required from OpenMS: https://github.com/OpenMS/OpenMS/issues/5164


src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2419,246 @@ namespace SmartPeak
...
+      return;
+    }
+
+    // determine type of spectral data (profile or centroided)
+    OpenMS::SpectrumSettings::SpectrumType spectrum_type = ms_peakmap[0].getType();
+
+    if (spectrum_type == OpenMS::SpectrumSettings::PROFILE)
+    {
+      //if (!getFlag_("force"))
+      if (false)
```
ChristopherAbram: I am not sure what happens here?

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2419,246 @@ namespace SmartPeak
...
+    if (feat_chromatograms.size() == feat_map.size())
+    {
+      for (size_t i = 0; i < feat_chromatograms.size(); ++i)
+      {
+        for (size_t j = 0; j < feat_chromatograms[i].size(); ++j)
+        {
+          rawDataHandler_IO.getExperiment().addChromatogram(feat_chromatograms[i][j]);
+        }
+      }
+    }
```
ChristopherAbram: Maybe it would be easier to iterate with range instead indexes here (but thats equivalent)

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2419,246 @@ namespace SmartPeak
...
+    else
+    {
+//      LOGE << "FF-Metabo: Internal error. The number of features (" << feat_chromatograms.size() << ") and chromatograms (" << feat_map.size() << ") are different! Aborting.";
+//      return;
```
ChristopherAbram: Should this be uncommented?

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2419,246 @@ namespace SmartPeak
...
+    }
+
+    // store ionization mode of spectra (useful for post-processing by AccurateMassSearch tool)
+    if (!feat_map.empty())
+    {
+      std::set<OpenMS::IonSource::Polarity> pols;
+      for (size_t i = 0; i < ms_peakmap.size(); ++i)
+      {
+        pols.insert(ms_peakmap[i].getInstrumentSettings().getPolarity());
+      }
```
ChristopherAbram: Here again, maybe range-based iteration would be better

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2419,246 @@ namespace SmartPeak
...
+    const ParameterSet& params_I,
+    const Filenames& filenames
+  ) const
+  {
+    LOGD << "START PickMS2Features";
+
+    if (params_I.count("PickMS2Features") && params_I.at("PickMS2Features").empty()) {
+      LOGE << "No parameters passed to PickMS2Features.";
+      LOGD << "END PickMS2Features";
+      return;
```
ChristopherAbram: I know that this approach is consistent with other process methods, however it seems to be hard for integration. No throw no return, no error flag, it seems to be imposible to programically detect error from caller point of view.

ChristopherAbram (commented):
In general I would have objections only to some parts of code (most optional). Maybe tests should be added, its much easier to add a test for someone who wrote the method (takes less time)

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: yes I agree.

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: yes it will be cleaned. this PR is actually a WIP

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: yes it will be cleaned. this PR is actually a WIP. by the way, we are failing there when I execute the workflow at the moment. so it needs to be investigated.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2423,231 @@ namespace SmartPeak
...
+    //-------------------------------------------------------------
+    FunctionParameters pick_ms2_feature_params;
+    if (params_I.count("PickMS2Features"))
+    {
+      pick_ms2_feature_params = params_I.at("PickMS2Features");
+    }
+
+    OpenMS::MassTraceDetection mtdet;
+    OpenMS::Param mtdet_parameters = mtdet.getParameters();
+    mtdet_parameters.remove("chrom_fwhm");
```
dmccloskey, line 2478: Is there a reason we need to remove this parameter entry?  I took a look at the `FeatureFinderMetabo.cpp` and Oliver does it there as well.  Would you mind asking him the reason for this over email or on Gitter?

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2423,231 @@ namespace SmartPeak
...
+
+    //-------------------------------------------------------------
+    // Setting input
+    //-------------------------------------------------------------
+    OpenMS::PeakMap ms_peakmap;
+    std::vector<int> ms_level(1, 1);
+    for (OpenMS::MSSpectrum& spec : rawDataHandler_IO.getExperiment().getSpectra()) {
+      ms_peakmap.addSpectrum(spec);
+    }
+    for (OpenMS::Chromatogram& chromatogram : rawDataHandler_IO.getExperiment().getChromatograms()) {
```
dmccloskey: From the warning below, it sounds like we should not add in chromatograms (in our case it should only be the TIC = total ion chromatogram which is a sum of all ions over time).  Can you confirm this?  If that is the case, the addition of chromatograms can be removed.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2423,231 @@ namespace SmartPeak
...
+
+    OpenMS::FeatureFindingMetabo ffmet;
+    OpenMS::Param ffmet_parameters = ffmet.getParameters();
+    Utilities::updateParameters(ffmet_parameters, params_I.at("MRMFeatureFinderScoring"));
+
+    //-------------------------------------------------------------
+    // Setting input
+    //-------------------------------------------------------------
+    OpenMS::PeakMap ms_peakmap;
+    std::vector<int> ms_level(1, 1);
```
dmccloskey: It does not look like this parameter is used anywhere.

dmccloskey: Do you know what these levels are supposed to indicate?  I have been looking through OpenMS and there appears to be no documentation on them anywhere.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2423,231 @@ namespace SmartPeak
...
+    }
+
+    //-------------------------------------------------------------
+    // configure and run feature finding
+    //-------------------------------------------------------------
+
+    OpenMS::FeatureMap feat_map;
+    std::vector< std::vector< OpenMS::MSChromatogram > > feat_chromatograms;
+    ffmet.setParameters(ffmet_parameters);
+    ffmet.run(m_traces_final, feat_map, feat_chromatograms);
```
dmccloskey, line 2594: We will want to update the MSExperiment chromatograms object with `feat_chromatograms` using e.g. calls to `rawDataHandler_IO.getExperiment().getChromatograms()`.  Note that we would want to add these chromatograms to the existing list and not overwrite it as there should only be 1 chromatogram (i.e., the TIC mentioned above) that we would want to preserve.

For visualization, we will want to check that the chromatogram has the attribute `getNativeID()` annotated.  We will want to the value of `getNativeID()` to match the `getUniqueID()` of the feature.  From looking at `FeatureFindingMetabo.cpp`, it looks like this is the case. 

Note that if the user sets the option for saving the convex hull, the points for the peak are stored in the feature which will also be used for visualization.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2423,231 @@ namespace SmartPeak
...
+      }
+    }
+
+    LOGI << "-- FF-Metabo stats --\n"
+      << "Input traces:    " << m_traces_final.size() << "\n"
+      << "Output features: " << feat_map.size() << " (total trace count: " << trace_count << ")";
+
+    // filter features with zero intensity (this can happen if the FWHM is zero (bc of overly skewed shape) and no peaks end up being summed up)
+    auto intensity_zero = [&](OpenMS::Feature& f) { return f.getIntensity() == 0; };
+    feat_map.erase(remove_if(feat_map.begin(), feat_map.end(), intensity_zero), feat_map.end());
```
dmccloskey: This could be problematic as the vectors for the `feat_map` and `feat_chromatograms` chromatograms could potentially become unaligned.  I would recommend configuring the algorithm to run over an e.g. std::pair consisting of `feat_map` and `feat_chromatograms` to keep the two vectors aligned.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2423,231 @@ namespace SmartPeak
...
+      }
+      feat_map[0].setMetaValue("scan_polarity", OpenMS::ListUtils::concatenate(sl_pols, ";"));
+    }
+
+    feat_map.setPrimaryMSRunPath({ rawDataHandler_IO.getMetaData().getFilename() });
+    LOGD << "setPrimaryMSRunPath: " << rawDataHandler_IO.getMetaData().getFilename();
+
+    rawDataHandler_IO.setFeatureMap(feat_map);
+    rawDataHandler_IO.updateFeatureMapHistory();
+
```
dmccloskey, line 2682: We will also want to updated the MSExperiment.Chromatograms as discussed above with the `feat_chromatograms`

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2423,231 @@ namespace SmartPeak
...
+    //-------------------------------------------------------------
+    // configure and run elution peak detection
+    //-------------------------------------------------------------
+    std::vector<OpenMS::MassTrace> m_traces_final;
+    Parameter* enable_elution = pick_ms2_feature_params.findParameter("enable_elution");
+    if (enable_elution && enable_elution->getValueAsString() == "true")
+    {
+      std::vector<OpenMS::MassTrace> splitted_mtraces;
+      epdet.setParameters(epdet_parameters);
+      // fill mass traces with smoothed data as well .. bad design..
```
dmccloskey, line 2555: This code is a bit of a mess...  That is quite impressive you were able to translate it given all of the mass spec jargon thrown in!  If I were in your situation, I am not sure if I would have been able to parse this so quickly.

src/tests/class_tests/smartpeak/source/RawDataProcessor_test.cpp
```diff
@@ -1345,6 +1347,84 @@ BOOST_AUTO_TEST_CASE(pickMS1Features)
...
+  PickMS2Features processor;
+
+  BOOST_CHECK_EQUAL(processor.getID(), -1);
+  BOOST_CHECK_EQUAL(processor.getName(), "PICK_MS2_FEATURES");
+}
+
+BOOST_AUTO_TEST_CASE(pickMS2Features)
+{
+  // Pre-requisites: load the parameters and associated raw data
+  ParameterSet params_1;
```
dmccloskey, line 1378: Can you also add in parameters for storing the convex hull and chromatograms by adding the associated options for `FeatureFinderMetabo`? From the documentation these are "report_convex_hulls" and "report_chromatograms". 

As a follow-up, can you then add tests for the convex hull data and additional chromatogram data?

dmccloskey (changes requested):
Nice work so far.  It looks like we will need to make some changes to the `SearchAccurateMass` object to update the chromatogram `NativeID` attribute for plotting with the updated names given to the features after calling `SearchAccurateMass`.  Let's discuss next week how best to do this.  It would probably also make sense to take a look at the issue for visualization (AUT-814) to make sure we don't make things harder than necessary when it comes to visualizing the data.

src/smartpeak/source/core/RawDataProcessor.cpp
dmccloskey: After quite a bit of searching was able to deduce that the 1 indicates that the spectra extracted should only be MS1.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2413,4 +2423,231 @@ namespace SmartPeak
...
+    OpenMS::Param ffmet_parameters = ffmet.getParameters();
+    Utilities::updateParameters(ffmet_parameters, params_I.at("MRMFeatureFinderScoring"));
+
+    //-------------------------------------------------------------
+    // Setting input
+    //-------------------------------------------------------------
+    OpenMS::PeakMap ms_peakmap;
+    std::vector<int> ms_level(1, 1);
+    for (OpenMS::MSSpectrum& spec : rawDataHandler_IO.getExperiment().getSpectra()) {
+      ms_peakmap.addSpectrum(spec);
```
dmccloskey: Given the comment above about wanting to extract out only MS1 level features, we will want to have a check in hear that queries the ms level (i.e., `getMSLevel()`) and only adds in spectra that are MS level 1.  The spectra should be a mix of both MS level 1 and 2.

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: yes, i will ask

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: yes, this parameter was used when loading the MzML file in the FeatureFinderMetabo example from OpenMS, to indicate we want only to load MS1 data (we will skip others).
obviously it is unused and will be removed
So we don't need to do equivalent from the rawDataHandler_IO ?

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: ok, I will remove this.

src/tests/class_tests/smartpeak/source/RawDataProcessor_test.cpp
bertrandboudaud: done.

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: done

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2484,19 +2484,15 @@ namespace SmartPeak
 
     OpenMS::FeatureFindingMetabo ffmet;
     OpenMS::Param ffmet_parameters = ffmet.getParameters();
-    Utilities::updateParameters(ffmet_parameters, params_I.at("MRMFeatureFinderScoring"));
```
dmccloskey: Sorry I didn't catch this previously

src/smartpeak/source/core/RawDataProcessor.cpp
dmccloskey: Please let me know if you need any additional clarifications on this.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2646,6 +2674,9 @@ namespace SmartPeak
 
     rawDataHandler_IO.setFeatureMap(feat_map);
     rawDataHandler_IO.updateFeatureMapHistory();
+    for (auto& chromatogram : merged_chromatograms) {
```
dmccloskey: I would recommend replacing this with `rawDataHandler_IO.getExperiment().setChromatograms(chromatogram);`.   I may have mispoken earlier today so I apologize if that was the case.

dmccloskey (changes requested):
Thanks for taking the time to speak earlier today.  This looks quite nice.  I left a couple of small comments.

dmccloskey (approved):
Looks great.

PR #317, opened 2021-02-15
[fix/aut826] Refresh parameters table only when parameters change

bertrandboudaud:
Refresh parameters table only when user parameters or workflow change

### DIFF
--- a/src/smartpeak/source/ui/ParametersTableWidget.cpp
+++ b/src/smartpeak/source/ui/ParametersTableWidget.cpp
@@ -1,0 +1,170 @@
+// --------------------------------------------------------------------------
+//   SmartPeak -- Fast and Accurate CE-, GC- and LC-MS(/MS) Data Processing
+// --------------------------------------------------------------------------
+// Copyright The SmartPeak Team -- Novo Nordisk Foundation 
+// Center for Biosustainability, Technical University of Denmark 2018-2021.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL ANY OF THE AUTHORS OR THE CONTRIBUTING
+// INSTITUTIONS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Douglas McCloskey $
+// $Authors: Douglas McCloskey, Pasquale Domenico Colaianni $
+// --------------------------------------------------------------------------
+
+#include <SmartPeak/core/ApplicationProcessor.h>
+#include <SmartPeak/ui/ParametersTableWidget.h>
+#include <algorithm>
+#include <map>
+#include <string>
+#include <utility>
+#include <vector>
+#include <imgui.h>
+
+namespace SmartPeak
+{
+
+  ParametersTableWidget::ParametersTableWidget(SessionHandler& session_handler, ApplicationHandler& application_handler, const std::string& table_id)
+    : session_handler_(session_handler), application_handler_(application_handler), table_id_(table_id)
+  {
+    application_handler_.addParametersObserver(this);
+    application_handler_.addWorkflowObserver(this);
+  };
+
+  ParametersTableWidget::~ParametersTableWidget()
+  {
+    application_handler_.removeParametersObserver(this);
+    application_handler_.removeWorkflowObserver(this);
+  }
+
+  void ParametersTableWidget::parametersUpdated()
+  {
+    refresh_needed_ = true;
+  }
+
+  void ParametersTableWidget::workflowUpdated()
+  {
+    refresh_needed_ = true;
+  }
+
+  void ParametersTableWidget::draw()
+  {
+    if (application_handler_.sequenceHandler_.getSequence().size() == 0) {
+      return;
+    }
+    std::vector<std::string> command_names = application_handler_.sequenceHandler_.getWorkflow();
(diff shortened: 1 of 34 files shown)


src/tests/class_tests/smartpeak/source/Parameters_test.cpp
```diff
@@ -345,6 +345,35 @@ BOOST_AUTO_TEST_CASE(ParameterSet_addFunctionParameters)
   BOOST_CHECK(param->getType() == "string");
 }
 
+BOOST_AUTO_TEST_CASE(ParameterSet_equal)
```
dmccloskey, line 348: Nice additional functionality

src/tests/class_tests/smartpeak/source/CastValue_test.cpp
```diff
@@ -383,29 +395,29 @@ BOOST_AUTO_TEST_CASE(castValue_comparison_operator)
...
 
+  BOOST_CHECK_EQUAL(c_bool_list1 == bool_list2, false);
   // not supported always return true
-  BOOST_CHECK_EQUAL(c_bool_list1 == bool_list2, true);
   BOOST_CHECK_EQUAL(c_bool_list1 >= bool_list2, true);
   BOOST_CHECK_EQUAL(c_bool_list1 > bool_list2, true);
   BOOST_CHECK_EQUAL(c_bool_list1 <= bool_list2, true);
   BOOST_CHECK_EQUAL(c_bool_list1 < bool_list2, true);
 
+  BOOST_CHECK_EQUAL(c_float_list1 == float_list2, false);
```
dmccloskey, line 405: Nice

src/smartpeak/source/ui/Widget.cpp
```diff
@@ -175,9 +176,20 @@ namespace SmartPeak
...
-
+    }
+    std::vector<std::string> command_names = application_handler_.sequenceHandler_.getWorkflow();
+    ParameterSet user_parameters = application_handler_.sequenceHandler_.getSequence().at(0).getRawData().getParameters();
+    if ((command_names != input_command_names_) || (user_parameters != input_user_parameters_))
+    {
+      BuildCommandsFromNames buildCommandsFromNames(application_handler_);
+      buildCommandsFromNames.names_ = command_names;
+      buildCommandsFromNames.process(); 
+      session_handler_.getParametersTable(user_parameters, buildCommandsFromNames.commands_, headers_, body_);
```
dmccloskey: This change moves the logic for when to decide when to update the parameters table to the Widget.  Do you think this is the strategy that we will move forward with in the future for the rest of the widgets?  My concern is that the widget is not just displaying the data but also deciding when to update the data.  Do you see any issues in regards to not separating the logic of updating the data and displaying the data?

dmccloskey (commented):
The additions are nice.  I would like to hear your thoughts on the comment I left about the combined scope for the widget class.

src/smartpeak/source/ui/Widget.cpp
bertrandboudaud: Yes, good point. The thing is when I implemented this, there was no way to know that the data, here, the couple (parameter, workflow) has changed. While I used workaround for the chromatogram (wait that the workflow finished to run, from GUI.cpp), here it's more complicated: 

1. Parameters can be loaded, maybe later edited, 
2. and more problematic there the workflow can be loaded, and edited. And we access directly the vector to push items to this.

To make it proper, we need to introduce observers to notify the view of the change from the model. The model needs means to know the data has changed (that is: not accessing vectors, or, give a mean to notify changes).

So I can proceed and introduce these changes with the opportunity of this PR. on this context, I see 2 options of doing that:

1. encapsulate the maps and vectors that we access directly (workflow, but in this case also Parameter fields) so that when the change occurs the view is notified. It could be a little bit too much. any change will update the view.
2.  introduce notification methods in the model (sequenceHandler). we still access to the vectors, maps, parameter fields, but the call will have to call the notification function from the model, when appropriate. It would reduce the notifications, giving better control on the update sequence, and maybe let open the option to do 1) later.
 so for example, for 2) we will do 
```
worflow_.push("MyCommand")
sequenceHandler.notifyWorkflowChange() // sessionhandler will notify observers (the views)
```

Obviously, we want to avoid View to view messages when updating. For example, we want to avoid to notify the parameter view from the workflow view when the workflow has been changed by the user. It would be the worst to do. this must come (or at least, pass-through) from the model.

dmccloskey: Please proceed with the solution that you think is best.  I agree that we need an overhaul of how and when we decide different views should be updated.  I think the `parameters` provides a good use case for beginning to develop such a solution/strategy.  If you would like to discuss the options or next steps further please just let me know.  Other than that I will wait until the next review request to continue our discussion after you have had time to test out and implement the options that you see as optimal.

src/tests/class_tests/smartpeak/source/WorkflowObservable_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Douglas McCloskey $
+// $Authors: Douglas McCloskey $
```
dmccloskey, line 21: Don't forget to add your name to these now.

src/smartpeak/source/ui/ParametersTableWidget.cpp
```diff
@@ -0,0 +1,170 @@
...
+    ImGui::SameLine();
+    ImGui::Checkbox("Show unused", &show_unused);
+
+    // headers
+    const ImGuiTableFlags table_flags = ImGuiTableFlags_Resizable | ImGuiTableFlags_Hideable | ImGuiTableFlags_Reorderable |
+      ImGuiTableFlags_NoBordersInBody | ImGuiTableFlags_Scroll | ImGuiTableFlags_SizingPolicyFixedX | ImGuiTableFlags_NoSavedSettings |
+      ImGuiTableFlags_Sortable | ImGuiTableFlags_MultiSortable;
+
+    // Columns we want to show up
+    std::vector<int> cols_to_show = { 0, 1, 2, 3, 7 };
```
dmccloskey, line 86: What columns do these correspond to?  I am just curious about the jump from 3 to 7 in the vector.

dmccloskey (commented):
This is very nice.  I see these changes as a big step towards where we want to see the application progress towards on the UI side.

I had some very small comments prior to merging.

src/smartpeak/source/ui/ParametersTableWidget.cpp
bertrandboudaud: The columns are these: "function","name","type","value","description","status","valid","restrictions","schema_type" "description","status","valid" are not displayed, but the description appears by hovering the mouse, and the 2 other columns are used to display color or error.

bertrandboudaud:
> This is very nice. I see these changes as a big step towards where we want to see the application progress towards on the UI side.
> 
> I had some very small comments prior to merging.

So as you see, I have done with the observer .. but I don't trigger for all modifications (we could have done notification in the encpasulation of workflow modification, or to the parameters modifications), for the moment, I think it should do the trick.

dmccloskey (approved):

PR #319, opened 2021-02-15
Feat/aut833 Configure changelog generation

ChristopherAbram:
This PR considers only configuration for automatically generated changelog.
📌 WIP: perhaps we will change commit message format (need your feeedback after tech meeting)


bertrandboudaud:
I would be interested in seeing a demo of that (the automatic changelog generation) before merging it.

dmccloskey (approved):
Definitely good to keep in our back pocket just in case.

PR #320, opened 2021-02-16
[feat/aut834] Include/Pass version tag to SmartPeak code.

ChristopherAbram:
+ Modify CMakeLists.txt to cache version information.
+ Add `Utilities::getSmartPeakVersion()` static method.
+ Include version information in GUI.cpp.

📌 I think this method doesn't need test.

### DIFF
--- a/src/examples/source/GUI.cpp
+++ b/src/examples/source/GUI.cpp
@@ -158,6 +158,9 @@ int main(int argc, char** argv)
     .addAppender(&consoleAppender)
     .addAppender(&appender_);
 
+  // Log SmartPeak launch initiated:
+  LOG_INFO << "Start SmartPeak version " << Utilities::getSmartPeakVersion();
+
   if (error_msg.empty())
   {
     if (logdir_created) LOG_DEBUG << "Log directory created: " << logdirpath;
@@ -177,6 +180,8 @@ int main(int argc, char** argv)
   }
 
   // Setup window
+  auto smartpeak_window_title = static_cast<std::ostringstream&&>(
+    std::ostringstream() << "SmartPeak v" << Utilities::getSmartPeakVersion()).str();
   SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
   SDL_GL_SetAttribute(SDL_GL_DEPTH_SIZE, 24);
   SDL_GL_SetAttribute(SDL_GL_STENCIL_SIZE, 8);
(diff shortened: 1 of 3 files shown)


src/examples/source/GUI.cpp
```diff
@@ -381,8 +381,9 @@ int main(int argc, char** argv)
     }
     if (ImGui::BeginPopupModal("About modal", NULL, ImGuiWindowFlags_AlwaysAutoResize))
     {
+      // TODO: read text from file and print in the window
```
bertrandboudaud, line 393: Would be better to add a task in JIRA than getting a TODO inside the code, for this kind of remark. TODO goes well for missing states handling, or identified performance issues while writing the code, for example. everything that would help someone debugging there. features should go to to JIRA. This is minor :)

bertrandboudaud (approved):
Looks good to me.
another place where the version should be displayed, is on the log, on startup, to ease debugging investigation based on the log.

CMakeLists.txt
```diff
@@ -78,9 +78,9 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Windows")
 #------------------------------------------------------------------------------
 # SmartPeak Version Information
 #------------------------------------------------------------------------------
-set (SMARTPEAK_PACKAGE_VERSION_MAJOR 1)
-set (SMARTPEAK_PACKAGE_VERSION_MINOR 0)
-set (SMARTPEAK_PACKAGE_VERSION_PATCH 0)
+set(SMARTPEAK_PACKAGE_VERSION_MAJOR "0" CACHE STRING "SmartPeak major version")
+set(SMARTPEAK_PACKAGE_VERSION_MINOR "0" CACHE STRING "SmartPeak minor version")
+set(SMARTPEAK_PACKAGE_VERSION_PATCH "0" CACHE STRING "SmartPeak patch version")
```
ghost: how are the version numbers carried out to the respective variables?

src/examples/source/GUI.cpp
ChristopherAbram: This TODO is more of a note for a future when we decide to make GUI modular - at the moment it can stay as it is

ChristopherAbram:
> Looks good to me.
> another place where the version should be displayed, is on the log, on startup, to ease debugging investigation based on the log.

Good point, I will append it to the log 👍

ghost (changes requested):

CMakeLists.txt
ChristopherAbram: I think we will pass these variables while lunching cmake config. By default version is `0.0.0` which means no version defined

ChristopherAbram: Please elaborate what you would like to modify

src/examples/source/GUI.cpp
```diff
@@ -158,6 +158,9 @@ int main(int argc, char** argv)
     .addAppender(&consoleAppender)
     .addAppender(&appender_);
 
+  // Log SmartPeak launch initiated:
+  LOG_INFO << "Start SmartPeak version " << Utilities::getSmartPeakVersion();
```
ChristopherAbram, line 162: I hope this satisfies your request @bertrandboudaud

CMakeLists.txt
ghost: > I think we will pass these variables while lunching cmake config. By default version is `0.0.0` which means no version defined

could you provide a link to the cmake config file you are referring to?:)

ChristopherAbram:
I have addressed @ahmedskhalil issue and also changed title bar for SmartPeak - now it also includes version.

ghost (approved):
very nice

PR #321, opened 2021-02-16
[feat/aut822] Description of the workflow steps displayed to the user

bertrandboudaud:
Description of commands are displayed to the user

1. The user will get the description when hovering the mouse on the workflow step (tooltip) from the workflow steps panel.
image

2. When adding a workflow step, the use will click on the add button and a pop up will appear where he can select the workflow to add. in this way he got access to the description before adding it. he can browse  the workflow steps and accept (add it) or cancel.
image

### DIFF
--- a/src/examples/source/GUI.cpp
+++ b/src/examples/source/GUI.cpp
@@ -227,6 +227,10 @@ int main(int argc, char** argv)
     // Intialize the window sizes
       win_size_and_pos.setXAndYSizes(io.DisplaySize.x, io.DisplaySize.y);
 
+      if ((!workflow_is_done_) && manager_.isWorkflowDone())
+      {
+        manager_.updateApplicationHandler(application_handler_);
+      }
       workflow_is_done_ = manager_.isWorkflowDone();
       file_loading_is_done_ = file_picker_.fileLoadingIsDone();
 
@@ -438,7 +442,7 @@ int main(int argc, char** argv)
             popup_file_picker_ = true;
             update_session_cache_ = true;
           }
-          if (ImGui::MenuItem("Workflow")) {
+          if (ImGui::MenuItem("Workflow", NULL, false, workflow_is_done_)) {
             static LoadSequenceWorkflow processor(application_handler_);
             file_picker_.setProcessor(processor);
             popup_file_picker_ = true;
@@ -784,6 +788,7 @@ int main(int argc, char** argv)
         }
         if (show_workflow_table && ImGui::BeginTabItem("Workflow", &show_workflow_table))
         {
+          workflow_.setEditable(workflow_is_done_);
           workflow_.draw();
           ImGui::EndTabItem();
         }
--- a/src/smartpeak/include/SmartPeak/core/ApplicationHandler.h
+++ b/src/smartpeak/include/SmartPeak/core/ApplicationHandler.h
@@ -92,6 +92,16 @@ namespace SmartPeak
           return sample_group_method->getParameterSchema();
       }
 
+      std::string getDescription() const
+      {
+        if (type == RawDataMethod)
+          return raw_data_method->getDescription();
+        else if (type == SequenceSegmentMethod)
+          return seq_seg_method->getDescription();
+        else
+          return sample_group_method->getDescription();
+      }
+
       std::shared_ptr<RawDataProcessor> raw_data_method;
       std::shared_ptr<SequenceSegmentProcessor> seq_seg_method;
       std::shared_ptr<SampleGroupProcessor> sample_group_method;
--- a/src/smartpeak/include/SmartPeak/core/ApplicationHandler.h
+++ b/src/smartpeak/include/SmartPeak/core/ApplicationHandler.h
@@ -77,7 +77,7 @@ namespace SmartPeak
       ParameterSet getParameterSchema() const
       {
         const auto description = getIProcessorDescription();
-        return (description ? description->getParameterSchema() : ParameterSet({}));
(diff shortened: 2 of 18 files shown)


src/smartpeak/include/SmartPeak/core/ApplicationHandler.h
```diff
@@ -92,6 +92,16 @@ namespace SmartPeak
           return sample_group_method->getParameterSchema();
       }
 
+      std::string getDescription() const
+      {
+        if (type == RawDataMethod)
+          return raw_data_method->getDescription();
+        else if (type == SequenceSegmentMethod)
+          return seq_seg_method->getDescription();
+        else
```
dmccloskey: We are scoped to add in at least one additional `type` in the future so I would probably recommend using an explicit `else if` instead of the `else` here.  I would also recommend some sort of error logging or warning if the type is not recognized.

dmccloskey (changes requested):
I think this will be very much appreciated by our users.  I know I will appreciate it!

I made a very small comment.  Other than that, I think it is ready to be merged.

bertrandboudaud:
ApplicationHandler has been fixed to handle the default common "else". For this, I took the opportunity to factorize code and use the IProcessorDescription. factorizing this, it will avoid repetition of code and will be less error prone. I also added getParameterSchema in the IProcessorDescription, which fits very well in this interface I think. (however i had little issue, at some point, sequenceProcessor needs a static decorated getParameterSchema  which I had renamed getParameterSchemaStatic to void collision)

bertrandboudaud:
By testing more, I saw some random crashes when the processing of the workflow ends. The reason is the std::move(appplication_handler) at the end of the running thread. this move is to allow the GUI to get fresh information.
However, nothing prevents GUI to access the application handler at this exact moment. And I don't think the rule "you should not access application_handler during processing" is assumed, nor reasonable - and this move, that we can consider atomic, even let think we can do it. 
However, in this PR, commands are instantiated to get the description, and doing this, some objects contained in the application handler are looped in. the std::move does not prevent thread-safety there. By the way, probably other parts of GUI, opened, updated at the wrong timing probably may crash also already. in this PR it's easily reproduced as we get the description at every frame.
To fix that, the std::move has been deported in the main thread, in GUI.cpp.

bertrandboudaud:
Added some check to prevent edition while workflow is running. It does not seems to involve a crash. i just saw some strange behavior like the edited workflow get reset to previous status when the workflow finished to run (because the swap of application_handler).

src/smartpeak/include/SmartPeak/core/ApplicationHandler.h
```diff
@@ -64,49 +64,49 @@ namespace SmartPeak
...
       int getID() const
       {
-        if (type == RawDataMethod)
-          return raw_data_method->getID();
-        else if (type == SequenceSegmentMethod)
-          return seq_seg_method->getID();
-        else
-          return sample_group_method->getID();
+        const auto description = getIProcessorDescription();
+        return (description ? description->getID() : 0);
```
dmccloskey, line 68: Nice

src/smartpeak/include/SmartPeak/iface/IProcessorDescription.h
```diff
@@ -20,6 +20,11 @@ namespace SmartPeak
     */
     virtual std::string getDescription() const = 0;
 
+    /**
+      Get the Schema (possible, default parameters) of the processor.
+    */
+    virtual ParameterSet getParameterSchema() const = 0;
```
dmccloskey, line 26: Nice

src/examples/source/GUI.cpp
```diff
@@ -784,6 +788,7 @@ int main(int argc, char** argv)
         }
         if (show_workflow_table && ImGui::BeginTabItem("Workflow", &show_workflow_table))
         {
+          workflow_.setEditable(workflow_is_done_);
```
dmccloskey, line 791: Makes sense

dmccloskey (approved):
> ApplicationHandler has been fixed to handle the default common "else". For this, I took the opportunity to factorize code and use the IProcessorDescription. factorizing this, it will avoid repetition of code and will be less error prone. I also added getParameterSchema in the IProcessorDescription, which fits very well in this interface I think. (however i had little issue, at some point, sequenceProcessor needs a static decorated getParameterSchema which I had renamed getParameterSchemaStatic to void collision)

I agree, it fits quite nicely.

> By testing more, I saw some random crashes when the processing of the workflow ends. The reason is the std::move(appplication_handler) at the end of the running thread. this move is to allow the GUI to get fresh information.
> However, nothing prevents GUI to access the application handler at this exact moment. And I don't think the rule "you should not access application_handler during processing" is assumed, nor reasonable - and this move, that we can consider atomic, even let think we can do it.
> However, in this PR, commands are instantiated to get the description, and doing this, some objects contained in the application handler are looped in. the std::move does not prevent thread-safety there. By the way, probably other parts of GUI, opened, updated at the wrong timing probably may crash also already. in this PR it's easily reproduced as we get the description at every frame.
> To fix that, the std::move has been deported in the main thread, in GUI.cpp.

It does appear as if this PR exposed a bug in the multi-threading logic.  That fix makes sense to me.

> Added some check to prevent edition while workflow is running. It does not seems to involve a crash. i just saw some strange behavior like the edited workflow get reset to previous status when the workflow finished to run (because the swap of application_handler).

Good catch on that.  These are hard to find unless you expose the application to a number of situations.

PR #322, opened 2021-02-19
[fix/aut815] - Updating GC-MS full scan workflow with MDV methods

ghost:
Objectives

- Updating `metadataToString` and `metadatafloatToString` with feature metadata for `average_accuracy` & `absolute_difference`.

### DIFF
--- a/src/smartpeak/include/SmartPeak/core/FeatureMetadata.h
+++ b/src/smartpeak/include/SmartPeak/core/FeatureMetadata.h
@@ -25,6 +25,7 @@
 
 #include <map>
 #include <string>
+#include <type_traits>
 
 namespace SmartPeak {
   enum class FeatureMetadata {
@@ -57,7 +58,8 @@ namespace SmartPeak {
     mz_error_ppm,
     mz_error_Da,
     average_accuracy,
-    absolute_difference/*,
+    absolute_difference,
+    SIZE_OF_FeatureMetadata/*,
     accuracy,
     n_features,
     validation*/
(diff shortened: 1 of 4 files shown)


dmccloskey (approved):
Everything looks good.  If you can double check that these feature metadata are also available to report when the user goes to "actions" -> "report" that would be great.

In regards to the build on Mac, it looks like there is a problem on the OpenMS side.

ghost:
> Everything looks good. If you can double check that these feature metadata are also available to report when the user goes to "actions" -> "report" that would be great.
> 
> In regards to the build on Mac, it looks like there is a problem on the OpenMS side.

Thanks for reminding, the new entries are now visible in the report's right pane.

src/smartpeak/include/SmartPeak/core/FeatureMetadata.h
```diff
@@ -67,5 +69,7 @@ namespace SmartPeak {
   extern const std::map<FeatureMetadata, std::string> metadataToString;
   extern const std::map<FeatureMetadata, std::string> metadatafloatToString;
 
-  constexpr size_t FeatureMetadataSize { 28 }; // was 31 with accuracy, n_features, and validation
+  constexpr size_t FeatureMetadataSize {
+    static_cast<std::underlying_type_t<FeatureMetadata>>(FeatureMetadata::SIZE_OF_FeatureMetadata) - 1
```
dmccloskey, line 73: That is quite cool!

PR #323, opened 2021-02-20
[feat/aut876] - Improvements/Changes to CI

ghost:
**Objectives**
- Changing builds on all target platforms to Release
- Annotating build steps
- Cleanup

**Future Improvements**
- Implementing interrupts when a step fails for `build_windows`
- Caching dependencies

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,8 +86,6 @@ jobs:
             brew remove yarn node
             brew install cmake qt5 sdl2
             brew install boost
-      - restore_cache:
-          key: contrib_build-cache-{{ checksum "contrib_build.tar.gz" }}
       - run:
           name: Cloning and building OpenMS
           command: |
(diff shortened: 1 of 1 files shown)


ChristopherAbram (approved):
Very nice update to label job stages, its more readable now.

If it comes to release build for CI, I was thinking to build all targets in Debug for all ci checks, and switch it to Release when release event triggered the pipeline (so we would have two modes of the same pipeline). At the moment it is good.

ghost:
> Very nice update to label job stages, its more readable now.
> 
> If it comes to release build for CI, I was thinking to build all targets in Debug for all ci checks, and switch it to Release when release event triggered the pipeline (so we would have two modes of the same pipeline). At the moment it is good.

I was also thinking the same , sometime we might need a debug version or RelWithDebInfo but these could be produced locally, so I am not certain if we need that

dmccloskey (approved):
Awesome.

PR #325, opened 2021-02-23
[fix/aut887] Fix 'HPLC UV Unknowns' preset

bertrandboudaud:
missing coma.

### DIFF
--- a/src/smartpeak/source/ui/Workflow.cpp
+++ b/src/smartpeak/source/ui/Workflow.cpp
@@ -78,7 +78,7 @@ namespace SmartPeak
                     "QUANTIFY_FEATURES",
                     "STORE_FEATURES" };
           else if (s_string == "HPLC UV Unknowns")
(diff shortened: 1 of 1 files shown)


src/smartpeak/source/ui/Workflow.cpp
```diff
@@ -78,7 +78,7 @@ namespace SmartPeak
                     "QUANTIFY_FEATURES",
                     "STORE_FEATURES" };
           else if (s_string == "HPLC UV Unknowns")
-            ids = { "LOAD_RAW_DATA"
+            ids = { "LOAD_RAW_DATA",
```
ChristopherAbram, line 81: Interesting bug, compiler in this case decides to concatenate two strings instead of failing...

ChristopherAbram (approved):
Good to go!

ghost (approved):
Looking good, macOS CI is working again on `develop`.

dmccloskey (approved):

PR #328, opened 2021-02-26
[Feat/aut800] - Switching to ImGui v1.81 Release

ghost:
**Objectives**

- Switching from the currently used ImGui version to ImGui v1.81 (latest release).
- Renaming `GUI` to `SmartPeakGUI`.


.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           name: Installing SmartPeak SuperBuild
           command: |
             cd ~/SmartPeak
-            #mkdir superbuild
+            mkdir -p superbuild
```
dmccloskey, line 42: Is this needed?  Now that we are pointing to a specific release we should not need to store ImGui in our repo correct?

.circleci/config.yml
```diff
@@ -124,11 +124,11 @@ jobs:
             ls -alt bin
             cpack -G DragNDrop && ls -alth
             yes | hdiutil attach -noautoopen -noverify SmartPeak-*.dmg > /dev/null
-            otool -l /Volumes/SmartPeak-1.0.0-Darwin/GUI.app/Contents/MacOS/GUI | grep -a name
-            cp -r /Volumes/SmartPeak-*-Darwin/GUI.app /Volumes/SmartPeak-*-Darwin/share .
-            xattr -l GUI.app/*
-            sudo xattr -dr com.apple.quarantine GUI.app
-            # open GUI.app
+            otool -l /Volumes/SmartPeak-1.0.0-Darwin/SmartPeakGUI.app/Contents/MacOS/SmartPeakGUI | grep -a name
```
dmccloskey, line 127: Nice

.circleci/config.yml
ghost: This is to make sure `superbuild` folder is newly created even if it exists so that using the latest ImGui pulled is used.

dmccloskey: I see.

dmccloskey (approved):

PR #337, opened 2021-03-17
Feat/aut795 - Make parameters editable

bertrandboudaud:
Make parameters editable.

* User can edit the parameter by clicking on the value of the parameter:
* User can edit the user-overridden parameter but also the default parameter (adding them to the user parameter), this why I use the "set" word for the button instead of ok, to emphasize that an override is done.
* User can also "unset" the parameter, removing the parameter from the list of user parameter (the workflow will then use default)

image

* string list constraint parameter, and boolean parameter are displayed with a combo box:
image
image

* constraints are verified when editing. that means, the "set" button will be disabled if the constraints are not verified
image

Parameters table is updated. but I think there is no way to export it at the moment.

from code point of view, It's a little bit sad to have to reconstruct a Parameter to pass it to the ParameterEditorWidget. would have been very convenient to get a list of Parameter as input list of the table widget instead of Tensors, but it works with some minor additional method to pass some information between the different structures.


src/smartpeak/source/core/Parameters.cpp
```diff
@@ -173,11 +173,72 @@ namespace SmartPeak
...
+        value_.s_ = value_as_string;
+      }
+    }
+  }
+
+  void Parameter::setRestrictionsFromString(const std::string& restriction_as_string)
+  {
+    if (restriction_as_string.empty())
+      return;
+    if ((restriction_as_string[0] == '[') && (restriction_as_string[restriction_as_string.size() - 1] == ']'))
```
dmccloskey: Is there a reason we cannot use the utilities function for converting from a string list here?

src/smartpeak/source/core/Parameters.cpp
```diff
@@ -173,11 +173,72 @@ namespace SmartPeak
...
+      {
+        auto c = std::make_shared<CastValue>();
+        Utilities::parseString(m_max[2], *c);
+        cast_max = c;
+      }
+      if (cast_min || cast_max)
+      {
+        setConstraintsMinMax(cast_min, cast_max);
+      }
+    }
```
dmccloskey: The code and logic here is very nice for pulling out the values, but I am curious if it is absolutely necessary?  Would it not be easier to instead create a method that e.g., takes in the `min`, `max`, or list parameters as arguments directly and then  applies the correct function accordingly?

src/smartpeak/source/ui/ParameterEditorWidget.cpp
```diff
@@ -0,0 +1,166 @@
...
+    {
+      for (const auto valid_string : valid_string_)
+      {
+        if (ImGui::Selectable(valid_string.c_str()))
+        {
+          std::fill(input_text_field_.begin(), input_text_field_.end(), 0);
+          if (valid_string.size() < input_text_field_.size())
+          {
+            std::copy(valid_string.begin(), valid_string.end(), input_text_field_.data());
+          } // else we are in trouble, abnormal long value parameter, do not display.
```
dmccloskey: Can this be logged?

src/smartpeak/source/ui/ParameterEditorWidget.cpp
```diff
@@ -0,0 +1,166 @@
...
+    {
+      // the parser will consider non decimal-annoted number as int, 
+      // which will not be valid if the parameter is float type.
+      // force conversion.
+      validity_test_cast = static_cast<float>(validity_test_cast.i_);
+    }
+    bool valid = parameter_.isValid(validity_test_cast);
+
+    if (!valid) ImGui::PushStyleVar(ImGuiStyleVar_Alpha, ImGui::GetStyle().Alpha * 0.5f);
+    if (ImGui::Button("Set") && valid)
```
dmccloskey: Nice choice of words here.

src/smartpeak/source/ui/ParameterEditorWidget.cpp
```diff
@@ -0,0 +1,166 @@
...
+    std::ostringstream default_value_stream;
+    default_value_stream << "Default value: " << default_value;
+    default_value_ = default_value_stream.str();
+    parameter_ = parameter;
+    std::string value = parameter.getValueAsString();
+    std::fill(input_text_field_.begin(), input_text_field_.end(), 0);
+    if (value.size() < input_text_field_.size())
+    {
+      std::copy(value.begin(), value.end(), input_text_field_.data());
+    } // else we are in trouble, abnormal long value parameter, do not display.
```
dmccloskey: You have taken great care to ensure that the user parameters are checked and validated.  Even so, I can imagine that there are edge cases or behavior that the user may experience that they might not understand that would be nice to debug quickly by inspecting the log.  Do you see any opportunities in the code here where some logging could be implemented to better help diagnose problems that the user (or developer) may find when change the parameters?

src/smartpeak/source/ui/ParametersTableWidget.cpp
```diff
@@ -164,6 +181,20 @@ namespace SmartPeak
         }
       }
       ImGui::EndTable();
+      if (edit_parameter)
+      {
+        // we rebuild the Parameter object from the selected row
```
dmccloskey: The Tensor format was chosen to enable 1) faster plotting and 2) faster data analytics.  None of those two apply here for the parameters so if it makes sense to use a different representation, I would highly recommend that.

src/tests/class_tests/smartpeak/source/Parameters_test.cpp
```diff
@@ -895,6 +925,46 @@ BOOST_AUTO_TEST_CASE(Parameter_isValid)
...
+  list_string->push_back("two");
+  list_string->push_back("three");
+  param_string.setConstraintsList(list_string);
+  std::vector<std::string> read_tags;
+  for (const auto& cast_value : param_string.getValidStrings())
+  {
+    read_tags.push_back(cast_value);
+  }
+  std::vector<std::string> expected{ "one", "two", "three" };
+  BOOST_CHECK_EQUAL_COLLECTIONS(read_tags.begin(), read_tags.end(), expected.begin(), expected.end());
```
dmccloskey: Nice.

dmccloskey (commented):
Very nice functionality that is well implemented!

My only major comment is on the logging and if there are areas of the code where logging could be added to make it easier to diagnose user problems down the road.

src/smartpeak/source/core/Parameters.cpp
bertrandboudaud: you're right. actually this is a side effect of getting the parameter table with tensors and therefore the restriction column Is a string. so at some part we need to parse back the string to set the value min and max, and also the list of string as above.
since the restriction column of the parameter table widget comes from the getRestrictionAsString() method from parameter i did this counter part, part of the Parameter class as well. which think makes sense (instead of parsing that string in the table widget and set the value parsed through a min and max method).
however, i see later you would recommend to use another representation for the table so may be we can have a try by getting a list of parameter instead of tensors, and hopefully all this method would be removed.

src/smartpeak/source/ui/ParameterEditorWidget.cpp
bertrandboudaud: yes, indeed. i will do that.

ghost (commented):
very nice. As discussed, adding a total parameter count plus another button to allow the user to "reset to defaults" can be beneficial to users.

ChristopherAbram (approved):
Good job! This feature has been eagerly requested and awaited by users.

bertrandboudaud:
I've done quite a large rework to get rid of the tensors for the parameter table. the parameter table just gets a list of parameters. in this way, most of the comments about parsing back strings are resolved now.

more logs are added in case the data is too long to be displayed (a message appears on the dialog box also). 

Based on the comments of Ahmed, I also added a "set to defaults" button, "remove" of unset and display the number of parameters on the table.

image

src/smartpeak/source/ui/ParameterEditorWidget.cpp
```diff
@@ -91,26 +87,41 @@ namespace SmartPeak
...
+          user_parameters.addParameter(function_parameter_, parameter_);
+        }
+        application_handler_.sequenceHandler_.notifyParametersChanged();
+        ImGui::CloseCurrentPopup();
       }
       else
       {
-        parameter_.setValueFromString(std::string(input_text_field_.data()), false);
-        user_parameters.addParameter(function_parameter_, parameter_);
+        LOGE << "ParameterEditorWidget : Parameter value for " << parameter_.getName() << ", " << input_text_field_.data();
```
dmccloskey, line 111: Nice

src/smartpeak/source/ui/ParameterEditorWidget.cpp
```diff
@@ -91,26 +87,41 @@ namespace SmartPeak
...
+        LOGE << "is invalid. Check constraints and type. Expected " << parameter_.getType() << " " << parameter_.getRestrictionsAsString();
       }
-      application_handler_.sequenceHandler_.notifyParametersChanged();
-      ImGui::CloseCurrentPopup();
     }
     if (!valid) ImGui::PopStyleVar();
 
     ImGui::SameLine();
-    if (ImGui::Button("Unset"))
+    if (ImGui::Button("Set to defaults"))
```
dmccloskey, line 118: I agree - "set to defaults" is a bit more intuitive.

src/smartpeak/source/ui/ParameterEditorWidget.cpp
```diff
@@ -163,4 +169,20 @@ namespace SmartPeak
...
+  {
+    std::fill(input_text_field_.begin(), input_text_field_.end(), 0);
+    if (value.size() < input_text_field_.size())
+    {
+      std::copy(value.begin(), value.end(), input_text_field_.data());
+    }
+    else
+    {
+      // else we are in trouble, abnormal long value parameter, do not display.
+      LOGE << "ParameterEditorWidget : Parameter value for " << parameter_.getName() << " is too long to display. (" << value.size() << ")";
```
dmccloskey, line 182: Nice

dmccloskey (approved):
Nice updates.

PR #338, opened 2021-03-17
Windows SmartPeak install: executable is not available from start menu

ChristopherAbram:
* Add cmake nsis configuration for creating application links,
* Add icon to executable on windows,
* Update cmake nsis config to include icon to installer and launch the application when installation complete,
* Configure default CPACK_GENERATOR for windows.


ChristopherAbram:
Single icon instead of folder:
image
Icon visible in windows taskbar:
image
Icon visible on the gui title bar:
image
Icon attached to installer as well:
image
Icon visible in `Control Panel\Programs\Programs and Features`
image

Issue: Installer still creates temporary directory in windows start menu, then the directory disapears...

bertrandboudaud:
It's very nice !! 

I'm not sure about
Issue: Installer still creates temporary directory in windows start menu, then the directory disappears... 
does not look a problem to me, we can live with that.

ChristopherAbram:
After some analysis I realized that I will leave the folder in start menu:
- nsis integration with cpack is badly documented and it was impossible to figure out how to disable uninstall icon (unistall icon is redundant), to be able to alter this default behaviour one would need to write custom nsis script (which is an overkill for now)
- perhaps in future we have more executables for front-end usage, therefore the folder will nicely group them.

dmccloskey (approved):
Looks good.

PR #341, opened 2021-03-23
Feat/aut828 - Added Statistics and more informations in the info panel

bertrandboudaud:
Originally we wanted to have a Dashboard. In the end, we split the dashboard into 2 places.
* The information panel will display some statistics.
* The graphs will be displayed in a dedicated statistics widget

image

The simple Info widget has been reworked to a dedicated widget.
Some colors are added to emphasize errors.
image


src/examples/source/SmartPeakGUI.cpp
```diff
@@ -172,13 +175,17 @@ int main(int argc, char** argv)
   GuiAppender appender_;
 
   // widgets
-  GenericTextWidget quickInfoText_;
+  InfoWidget quickInfoText_;
```
dmccloskey, line 178: Should have been done a while ago.  Thanks for making this a formal widget now.

dmccloskey (approved):
Great work.  Let's get this into the hands of our users and see what they think.

PR #348, opened 2021-03-31
[fix/aut780] FIAMS_FullScan_Unknown_test.exe assertion failure

ChristopherAbram:
Seems like for some reason `f1->getIntensity()` and `f1->getMetaValue("peak_apex_int")` return an empty value, even though corresponding metadata are set in XML file. Also, `fm1.size()` returns 1225 and file contains `3161` features.
The code `(double)f1->getMetaValue("peak_apex_int")` causes following exception:
```bash
---------------------------------------------------
FATAL: uncaught exception!
---------------------------------------------------
last entry in the exception handler: 
exception of type ConversionError occurred in line 478, function OpenMS::DataValue::operator double() const of /home/krzysztof/Workspace/OpenMS/src/openms/source/DATASTRUCTURES/DataValue.cpp
error message: Could not convert DataValue::EMPTY to double
---------------------------------------------------
```

The behaviour for f2 is correct for all cases.

📌 [WIP]: I need some input from the team on this issue.


src/examples/source/FIAMS_FullScan_Unknown_test.cpp
```diff
@@ -56,13 +57,13 @@ void test_main_FIAMS_FullScan_Unknown()
   std::cout << "0 calc_mass_to_charge: " << a1->calc_mass_to_charge.get() << std::endl;
   std::cout << "0 chemical_formula: " << a1->chemical_formula.get() << std::endl;
   std::cout << "0 identifier: " << a1->identifier.get().at(0).get() << std::endl;
-  assert(Utilities::assert_close((double)f1->getMetaValue("peak_apex_int"), 3.5241834375e05));
+  // assert(Utilities::assert_close((double)f1->getMetaValue("peak_apex_int"), 3.5241834375e05));
```
dmccloskey: Is this value empty?  It should not be.

src/examples/source/FIAMS_FullScan_Unknown_test.cpp
```diff
@@ -41,11 +41,12 @@ void test_main_FIAMS_FullScan_Unknown()
   
   std::cout << "fm1 size: " << fm1.size() << std::endl;
   std::cout << "fm2 size: " << fm2.size() << std::endl;
-  assert(fm1.size() == fm2.size());
```
dmccloskey: This is a problem.  The two sizes should be equal.

src/examples/source/FIAMS_FullScan_Unknown_test.cpp
```diff
@@ -56,13 +57,13 @@ void test_main_FIAMS_FullScan_Unknown()
...
   std::cout << "0 identifier: " << a1->identifier.get().at(0).get() << std::endl;
-  assert(Utilities::assert_close((double)f1->getMetaValue("peak_apex_int"), 3.5241834375e05));
+  // assert(Utilities::assert_close((double)f1->getMetaValue("peak_apex_int"), 3.5241834375e05));
   assert(Utilities::assert_close((double)f1->getRT(), 0.0));
   assert(Utilities::assert_close((double)a1->calc_mass_to_charge.get(), 102.957));
   assert(static_cast<std::string>(a1->chemical_formula.get()) == "H3O3P");
   assert(a1->identifier.get().at(0).get() == "HMDB:HMDB0001443");
-  assert(Utilities::assert_close((double)f1->getIntensity(), (double)f2->getIntensity()));
-  assert(Utilities::assert_close((double)f1->getMetaValue("peak_apex_int"), (double)f2->getMetaValue("peak_apex_int")));
+  // assert(Utilities::assert_close((double)f1->getIntensity(), (double)f2->getIntensity()));
```
dmccloskey: These values should be equal.

dmccloskey (changes requested):
Can investigate further why the generated and "test" featureMap are no longer the same?  I would recommend starting with where they are divergent by printing out the names of the features and seeing where the discrepencies begin.

src/examples/source/FIAMS_FullScan_Unknown_test.cpp
```diff
@@ -45,49 +45,56 @@ void test_main_FIAMS_FullScan_Unknown()
...
-  assert(Utilities::assert_close((double)f1->getIntensity(), (double)f2->getIntensity()));
-  assert(Utilities::assert_close((double)f1->getMetaValue("peak_apex_int"), (double)f2->getMetaValue("peak_apex_int")));
-  assert(Utilities::assert_close((double)f1->getRT(), (double)f2->getRT()));
-  assert(Utilities::assert_close((double)a1->calc_mass_to_charge.get(), (double)a2->calc_mass_to_charge.get()));
-  assert(static_cast<std::string>(a1->chemical_formula.get()) == static_cast<std::string>(a2->chemical_formula.get()));
-  assert(a1->identifier.get().at(0).get() == a2->identifier.get().at(0).get());
+  // First feature
+  {
+    const auto& f1 = fm1[0].getSubordinates()[0];
+    const auto& f2 = fm2[0].getSubordinates()[0];
```
ChristopherAbram, line 53: `getSubordinates()` is required to read correct values of meta data such as `peak_apex_int`.
I was inspired by this: https://github.com/OpenMS/OpenMS/blob/d9692da0d410c06b6cdc960f608a5c962360d09c/src/tests/class_tests/openms/source/MRMFeatureSelector_test.cpp#L97

dmccloskey (commented):
Thanks for making the updates.

dmccloskey (approved):

PR #349, opened 2021-04-01
Feat/aut754 Refactor of Feature line and heatmap plots

bertrandboudaud:
**Improved Heatmap:**
* Added color scale
* Display values on the grid
* Only one feature is displayed at a time (accessible through combo box)
* Details are displayed on tooltip when cell is hovered
* Handled different case of error: invalid values, "not enough data"

image

**Plot lines**
* the grid match the injection number column
* no more tick on the horizontal axis
* some padding around the plot
* tooltip when you hover the plot.

image


src/examples/source/SmartPeakGUI.cpp
```diff
@@ -1186,20 +1190,37 @@ int main(int argc, char** argv)
...
+          if (!heatmap_plot_widget)
+          {
+            heatmap_plot_widget = std::make_unique<Heatmap2DWidget>(
+              session_handler_,
+              application_handler_.sequenceHandler_,
+              "Heatmap Main Window");
+          }
+          // To be replace when we have implement events
+          static bool heatmap_initialized = false;
+          if (!workflow_is_done_)
```
dmccloskey, line 1210: If I understand this correctly, the user would not be able to view the heatmap while the workflow is running?  Is this restriction needed?

src/smartpeak/include/SmartPeak/core/SessionHandler.h
```diff
@@ -136,7 +136,29 @@ namespace SmartPeak
                                 const std::set<std::string>& scan_names,
                                 const std::set<std::string>& component_group_names);
     void setFeatureLinePlot();
-    void setFeatureHeatMap();
+
```
dmccloskey, line 139: Looks much cleaner

src/smartpeak/source/core/SessionHandler.cpp
```diff
@@ -1564,28 +1564,27 @@ namespace SmartPeak
           if (!selected_transition_groups(i).empty())
             component_group_names.insert(selected_transition_groups(i));
         }
-        // Update teh unique transitions
+        // Update the unique transitions
```
dmccloskey, line 1567: Ooops :)

src/smartpeak/source/ui/Heatmap2DWidget.cpp
```diff
@@ -6,14 +6,169 @@ namespace SmartPeak
...
+      {
+        session_handler_.getHeatMap(sequence_handler_, heatmap_data_, selected_feature_);
+        // We need to handle "invalid" data - infinite values, or vey high - which will crash ImPlot.
+        invalid_data_ = false;
+        for (int i = 0; i < heatmap_data_.feat_heatmap_row_labels.size(); ++i)
+        {
+          for (int j = 0; j < heatmap_data_.feat_heatmap_col_labels.size(); ++j)
+          {
+            int index = i * heatmap_data_.feat_heatmap_row_labels.size() + j;
+            if (heatmap_data_.feat_heatmap_data.data()[index] > 1e20)
```
dmccloskey: If 1e20 is the limit you found for the high value data, I would recommend making this a global constraint that can be shared among all of the plotting widgets.  Perhaps as a base class member?

src/smartpeak/source/ui/Heatmap2DWidget.cpp
```diff
@@ -6,14 +6,169 @@ namespace SmartPeak
...
+      
+      if (invalid_data_)
+      {
+        ImGui::Text("Some data for this feature seem to be Invalid (unexpected high), cannot display Heatmap");
+      }
+      else
+      {
+        if (heatmap_data_.feat_heatmap_row_labels.size() > 0 && heatmap_data_.feat_heatmap_col_labels.size() > 0)
+        {
+          assert(heatmap_data_.feat_heatmap_data.dimension(0) == heatmap_data_.feat_heatmap_row_labels.size()
```
dmccloskey: Would it be possible to replace the `assert` with a conditional statement that avoids plotting data instead of stopping the program?  I would also recommend an additional log of the violation.

src/smartpeak/source/ui/Widget.cpp
```diff
@@ -298,16 +298,81 @@ namespace SmartPeak
...
+              if (tooltip_exists)
+              {
+                ImGui::Separator();
+              }
+              std::ostringstream os;
+              os << "Injection: " << x_labels_(injection_number);
+              ImGui::Text(os.str().c_str());
+              os.str("");
+              os.clear();
+              os << "Serie: " << series_names_(i);
```
dmccloskey:
```suggestion
              os << "Series: " << series_names_(i);
```

src/tests/class_tests/smartpeak/source/SessionHandler_test.cpp
```diff
@@ -688,5 +688,13 @@ BOOST_AUTO_TEST_CASE(setCalibratorsScatterLinePlot1)
   SessionHandler session_handler;
   session_handler.setCalibratorsScatterLinePlot(testData.sequenceHandler);
 }
+BOOST_AUTO_TEST_CASE(getHeatMap)
+{
+  TestData testData;
+  SessionHandler session_handler;
+  SessionHandler::HeatMapData result;
+  std::string feature_name = "calculated_concentration";
+  session_handler.getHeatMap(testData.sequenceHandler, result, feature_name);
```
dmccloskey, line 697: Should there be test statements for the properties of the struct `result`?

dmccloskey (changes requested):
Very nice upgrade of these plots.  The refactors to the underlying code base are much welcomed as well.  I am starting to see the ground work being laid towards the event system.  The modularizations of the widgets for the GUI and refactors to allow for that are great to see as well.

I only had some small comments and questions.

src/examples/source/SmartPeakGUI.cpp
bertrandboudaud: The heatmap can be displayed, user can open and display the heatmap during the run. but if it is displayed during a running workflow, the data will be refreshed only when the workflow will be finished. said another way, it's just a means to notify the heatmap that the data has changed, do it can be updated automatically.

src/smartpeak/source/ui/Heatmap2DWidget.cpp
bertrandboudaud: yes, it's a value id decided on my side really, by experiment. I will move it to some constant.

bertrandboudaud:
Request modifications applied.

dmccloskey (approved):

PR #350, opened 2021-04-09
CI: use requirements.txt to install python dependencies

bertrandboudaud:
CI failed on the trigger release step.

https://app.circleci.com/pipelines/github/AutoFlowResearch/SmartPeak?branch=develop
(this release trigger has been triggered because there was the word "release" in the comments right?)

it seems to be because we don't get python dependence modules with the requirment.txt

image


ChristopherAbram (approved):
Good catch

PR #357, opened 2021-04-25
Fixed compilation with OpenMS

bertrandboudaud:
OpenMS has changed the way they implement list in the Parameters. They now use standard lists.


dmccloskey (approved):
This is quite good that they got rid of the StringList.  Next steps will be to get rid of all of the unnecessary calls to QT.

PR #359, opened 2021-04-26
[fix/aut945] Chromatogram/Spectrum: Handled too much points to display…

bertrandboudaud:
[fix/aut945] Chromatogram/Spectrum: Handled too many points to display error

PointsOverflow

instead of returning success, the session handler method set the error in the result data structure, which makes it more convenient to use from the widgets.

Also the number of point as been decreased (it was 9000 but I had the exception from around 6800

In this PR also, c++ 17 is enabled since OpenMS now requires it


ChristopherAbram (approved):
Seems that OpenMS requires now C++17, I think it looks okay

bertrandboudaud:
> Seems that OpenMS requires now C++17, I think it looks okay

yes, compiles now, and just in case i launched and test the app. no problem

PR #363, opened 2021-05-03
Feat/aut165 - Ability to edit sequence

bertrandboudaud:
* Ability to edit sample group
* Ability to edit sequence segment
* Ability to edit sample type
* Ability to do the above operation on multiple cells, using usual shift and control key modifier.

Widget has been extended to open some method to be overridden by specialized class to allow editing:
- isEditable()
- onEdit()

**Notes:**

1. The original Ticket also lists the possibility to change the tags but I don't see tags in that window.
2. @dmccloskey you told me to add for replicate_group_name column  but I don't see that columns in that window. by the way I don't see replicate_group_name column in our examples files, but i find it in some example even by our users. So I can add this column, we need to add code, in session handler. i can do it now.
3. Still, it's a bit weird to get the data properly updated. I introduced a MainGuiObserver that refresh the table of the sessionhandler. It's still not very consistent, but hopefully it will be soon when we will implement the events.

SequenceEditor


src/smartpeak/source/ui/SequenceGroupsEditorWidget.cpp
```diff
@@ -0,0 +1,122 @@
...
+          {
+            current_choice_ = valid_string;
+          }
+        }
+        ImGui::EndCombo();
+      }
+    }
+
+    static const std::vector<char> forbiden_chars = { ';',',' };
+    bool new_group_name_is_valid = !std::any_of(forbiden_chars.begin(), forbiden_chars.end(),
```
dmccloskey, line 80: Nice check.

dmccloskey (commented):
Overall, very nice.  Please ignore my comment about `tags`.  It is no longer relevant.

If you don't mind adding in the necessary code and checks for the `replicate_group_name` that would be great.  It is used by many of the downstream BFAIR workflows and will be needed for an eventual addition of `replicateGroupHandler` and `replicateGroupProcessor` for performing different statistical analyses and unsupervised ML algorithms on the data that are quite useful for debugging the dataset.  I leave that to your discretion whether that should be implemented in this PR or in a new one.

My only other commend would be if there are any opportunities for introducing additional unit tests into the methods you have added.  I see many of the depend upon the ImGui context, but some of them appear to only depend on class members.  I would be interested to hear your thoughts if unit testing would be useful for some of the methods introduced.

src/smartpeak/source/ui/SequenceGroupsEditorWidget.cpp
bertrandboudaud: by the way, do we allow one of these editable fields to be empty (I guess at least the replicate_group_name will be)?

dmccloskey: They can be empty strings.  That could serve as the "default" grouping.

bertrandboudaud:
> Overall, very nice. Please ignore my comment about `tags`. It is no longer relevant.
> 
> If you don't mind adding in the necessary code and checks for the `replicate_group_name` that would be great. It is used by many of the downstream BFAIR workflows and will be needed for an eventual addition of `replicateGroupHandler` and `replicateGroupProcessor` for performing different statistical analyses and unsupervised ML algorithms on the data that are quite useful for debugging the dataset. I leave that to your discretion whether that should be implemented in this PR or in a new one.
> 
> My only other commend would be if there are any opportunities for introducing additional unit tests into the methods you have added. I see many of the depend upon the ImGui context, but some of them appear to only depend on class members. I would be interested to hear your thoughts if unit testing would be useful for some of the methods introduced.

I added support for replicate_group (optional, to be backward compatible) and added some tests. 
About unit tests, some functions you may have detected have been moved private/protected, and I don't think we should test private/protected methods.

src/smartpeak/source/io/SequenceParser.cpp
```diff
@@ -275,6 +211,51 @@ namespace SmartPeak
         throw;
       }
     }
+  };
+
+  void SequenceParser::readSequenceFile(
```
dmccloskey, line 216: Nice modularization.

dmccloskey (approved):
Looks good on my end.  Nice work!

PR #364, opened 2021-05-04
[fix/aut942] Fixed crash on Linux when showing transition tab

bertrandboudaud:
The root cause of the crash was that the function lexicographical_sort was not returning any values in some cases (when value was equal, or partially equal), which leads to undefined behavior (we actually got random values on windows and a crash on Linux)

The PR includes unit tests to cover these cases.


dmccloskey (approved):
Good catch.  Thanks for taking care of this.  Unfortunate unit testing was not in place for those functions previously...

PR #365, opened 2021-05-05
SmartPeak headless mode

ChristopherAbram:
First version of `SmartPeak CLI` almost complete. I am submitting this PR for your initial review.
Things considered as done:
- Load session,
- Run integrity checks,
- Run workflow,
- Modify report options,
- Export reports,
- Specify an output path,
- A few extra options (coloring console etc.)
- User documentation - readthedocs,
- In code documentation,
- Quite good test coverage.

Note: 
- Passing workflow parameters is still in progress.

Screenshot from docs:
image


ChristopherAbram:
@dmccloskey It would be great to build readthedocs for this branch, I've added some more conent and I wonder whether it renders correctly. (You have an access!)

docs/about/parameters.rst
```diff
@@ -0,0 +1,279 @@
+SmartPeak Parameters
```
bertrandboudaud, line 1: It's sad that this cannot be generated easily :sweat_smile:  we have the information spread in the code, but no easy solution.

src/smartpeak/include/SmartPeak/core/ConsoleHandler.h
```diff
@@ -0,0 +1,180 @@
...
+     * 
+     * @see use_colors()
+     * @param init_file_appender set to true if location is correct, false otherwise
+    */
+    virtual void _initialize_log_console(bool init_file_appender = true);
+
+private:
+    /* ctor/dtor */
+    ConsoleHandler() 
+        : m_severity {plog::info}, m_enable_colors {false},
```
bertrandboudaud, line 161: do you need to initialize m_logdir_created?

bertrandboudaud (approved):
This is very nice work there !!  :smiley:

It is sad, that this complete, and detailed documentation has some redundant information that we have in other places (in the code). we will have to maintain this documentation along with the modifications of the code. but there is no easy solution.

As I tried to explain last sprint meeting, would be nice to be able to override workflow parameters, (in another task), such as the number of running threads, etc (ideally we would not hardcode a list but get it from the parameters that we load along with the workflow).

ghost (approved):

PR #367, opened 2021-05-18
Feat/aut924 - Progress information

bertrandboudaud:
* Display a progress bar, estimated time, and more detailed information about the current status of a running workflow.
  * The estimated time is very rough. It has a good result when using a unique type of commands in the workflow, but it becomes very rough when mixing types (like FIAMS workflow). A tooltip warns the user about that. Note that this estimation could be improved, by either working more on this, and more probably by using some weight values associated with the steps and amount of work to do for each.
  * The progress bar however shows the exact executed steps.
  
Progress

* Generalized usage of the observer along with an "event dispatcher" to avoid threading synchronization issues (the event dispatcher store the event coming from the observers and can dispatch them to the GUI thread).
  * removed some code in the GUI.cpp that refresh data. now use the observers as much as possible.

* ProgressInfo class provides information about the current status of the running workflow. It's a non-gui class that can be reused in the client interface.


src/examples/source/SmartPeakGUI.cpp
```diff
@@ -351,62 +348,24 @@ int main(int argc, char** argv)
...
 
       // Make the quick info text
-      quickInfoText_->setWorkflowDone(workflow_is_done_);
       quickInfoText_->setFileLoadingDone(file_loading_is_done_, file_picker_.errorLoadingFile());
       quickInfoText_->clearErrorMessages();
       if (exceeding_plot_points_) quickInfoText_->addErrorMessage("Plot rendering limit reached.  Not plotting all selected data.");
       if (exceeding_table_size_) quickInfoText_->addErrorMessage("Table rendering limit reached.  Not showing all selected data.");
       if (ran_integrity_check_ && integrity_check_failed_) quickInfoText_->addErrorMessage("Integrity check failed.  Check the `Information` log.");
       if (ran_integrity_check_ && !integrity_check_failed_) quickInfoText_->addErrorMessage("Integrity check passed.");
-      if (workflow_just_finished)
```
dmccloskey, line 371: Nice to see this is no longer needed.

src/examples/source/SmartPeakGUI.cpp
```diff
@@ -470,109 +429,91 @@ int main(int argc, char** argv)
             static LoadSequenceTransitions processor(application_handler_);
             file_picker_.setProcessor(processor);
             file_picker_.visible_ = true;
-            update_session_cache_ = true;
```
dmccloskey, line 473: Same here.

src/smartpeak/include/SmartPeak/core/EventDispatcher.h
```diff
@@ -0,0 +1,118 @@
...
+    SampleGroupProcessorObservable,
+    SequenceObservable,
+    TransitionsObservable
+  {
+    EventDispatcher() = default;
+
+    /**
+      IApplicationProcessorObserver
+    */
+    virtual void onApplicationProcessorStart(const std::vector<std::string>& commands) override;
```
dmccloskey, line 70: Does it make sense to break these into different classes?

src/smartpeak/include/SmartPeak/core/ProgressInfo.h
```diff
@@ -0,0 +1,170 @@
...
+      application_processor_observable.addApplicationProcessorObserver(this);
+      sequence_processor_observable.addSequenceProcessorObserver(this);
+      sequence_segment_processor_observable.addSequenceSegmentProcessorObserver(this);
+      sample_group_processor_observable.addSampleGroupProcessorObserver(this);
+    };
+
+    /**
+      IApplicationProcessorObserver
+    */
+    virtual void onApplicationProcessorStart(const std::vector<std::string>& commands) override;
```
dmccloskey, line 69: I am thinking this class could also potentially be subdivided further.

src/smartpeak/include/SmartPeak/core/ProgressInfo.h
```diff
@@ -0,0 +1,170 @@
...
+    virtual void onSampleGroupProcessorSampleStart(const std::string& segment_name) override;
+    virtual void onSampleGroupProcessorSampleEnd(const std::string& segment_name) override;
+    virtual void onSampleGroupProcessorEnd() override;
+  
+  public:
+
+    /**
+      @brief returns wether the workflow is running 
+    */
+    bool isRunning() const { return running_; }
```
dmccloskey, line 103: These functionalities could potentially be included in a base class.

src/smartpeak/source/core/ProgessInfo.cpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+      return std::chrono::steady_clock::duration::zero();
+    }
+  }
+
+  std::optional<std::chrono::steady_clock::duration> ProgressInfo::estimatedRemainingTime() const
+  {
+    if (running_batch_  && running_batch_->current_step_ > 0)
+    {
+      auto time_per_steps = ((running_batch_->last_item_end_time_point_ - application_processor_start_time_) / running_batch_->current_step_);
```
dmccloskey, line 142: I see this as requiring a lot of work to estimate more accurately given the complexity of workflow execution.  One idea would be to provide some sort of "time_factor" for each processor that we could scale upon an initial estimate of the system performance.  However, let's wait to see how well the current implementation is received by our users and if the time and effort is worth pursuing.

dmccloskey (commented):
Very nice work!  I have not gone through the code in extreme detail (which would take quite a bit of time and I don't want to delay getting this tested by our users).  I left a couple of high level comments that would be great to get your thoughts on before merging.

src/smartpeak/source/core/ProgessInfo.cpp
bertrandboudaud: yes, we can do better :)

src/smartpeak/include/SmartPeak/core/EventDispatcher.h
bertrandboudaud: I don't see the need to split into different classes, do you have specific in mind (a concern?). having a unique dispatcher looks correct, and also in this way the event are stored and dispatched in order.

src/smartpeak/include/SmartPeak/core/ProgressInfo.h
bertrandboudaud: In the same way, I don't see the need to subdivide. it's a lot of observer subscriptions, but the progress info needs to gather all this information to compute progress info and to present info to the user. Please let me know what is your concern your if you think about another solution.

src/smartpeak/include/SmartPeak/core/ProgressInfo.h
bertrandboudaud: Indeed this one isRunning may be an issue in design, as this information is already available through the workflow. i can rework there. I was thinking however it's quite convenient and this information maybe part of a class that provide information about the status of a running workflow. in this way, a class that needs to display some information to the user about a the running workflow doesn't need a direct dependency to the workflow.
About a base class it's possible, could you elaborate a little bit what you have in mind? to be sure we are aligned?

src/smartpeak/include/SmartPeak/core/ProgressInfo.h
bertrandboudaud: Ah, maybe you mean that we should have an interface in that class that only provide the information and where all these observer callback does not appear?

ProgressInfoBase     <-- base classe used by the GUI
 * getRunningTime()

ProgressInfoImpl : public ProgressInfoBase, public ApplicationProcessorObserver
 * onApplicationProcessorStart(..)

src/smartpeak/source/core/ProgessInfo.cpp
bertrandboudaud: time_factor (some kind of weight) could be a solution indeed. we may need some metrics as well. there are ways let see how it would be useful for the users!

src/smartpeak/include/SmartPeak/core/EventDispatcher.h
dmccloskey: From my reading of the code, it seemed that several of the methods only correspond to specific processors, so it might make sense to make processor specific classes that inherit common methods from a base class.

src/smartpeak/include/SmartPeak/core/ProgressInfo.h
dmccloskey: More in line with my comment above about the processor-specific methods.  Essentially we would have a common base class with methods such as this and then processor specific classes that only have the methods intended for those processors.  It is not a huge issue in my opinion, but wanted to get your thoughts.

src/smartpeak/source/core/ProgessInfo.cpp
dmccloskey: Very true.  Let's not spend too much time on this as we have not validated what is most useful to the users at this stage yet.

dmccloskey (approved):
The code looks good from my end.  Unless you see anything actionable on my comments, I would be in favor of merging.

src/smartpeak/include/SmartPeak/core/ProgressInfo.h
bertrandboudaud: ok, i think we can go with this current solution for the moment. in a way your concern is about single responsibility principle, which is from another point of view is respected as for this class the supposed responsibility is to give the big picture of the progress information and need to gather all these information coming from the observers. if one day we need to provide partial information (if we need to use the event dispatcher to the progress info class for only one piece of partial information), we may rework it. i am for merging in that way now and see how it goes. I don't think it's going to be bad fondation as long as we do not diverge on the responsibility of this class in the future.

PR #371, opened 2021-06-01
Feat/aut952 - reworked FilePicker handlers

bertrandboudaud:
A new interface, IFilePickerHandler, aims at handling action when the user has selected a file from the FilePicker.

image

* It reduces redundancy (in the end, we could even go further about refactorizing code).
* reduce classes declaration, no need to create a new class (where most of them were dummy classes) to handle the action, it was - to my point of view annoying and confusing to split code in multiple classes and files just to support file selection. only implement IFilePickerHandler pure virtual method onFilePicked, on processors (or any other class) that may handle file picking.
* handlers are not necessarily processors anymore.
* what was processor and now FilePickerHandlers have been dispatched into respective classes (SequenceParser, RawDataProcessor...)
* ApplicationProcessor now holds only classes that relate directly to the "Application's life cycle".

This PR also implement:
AUT-808 Export commands: display FilePicker dialog, for
* Sequence Analyst
* Sequence Hunter
* Sequence Xcalibur

This PR also Fix an issue where when storing file (save as ..), it was necessary to select one existing file from the FilePicker.

Some known issues (possible improvements)
* little bit weird to pass the application handler on the onFilePicker() method. It would be required to be able to pass that when instantiating the underlying class (processor, or whatever). it would requires lot of rework and a review of how the processors are instantiated.
* Would be nice to add a defaultFileName() on the IFilePickerHandler so that when the FilePickers displays, it is already filled with the good value (would be specially useful for the storing handlers).


dmccloskey (approved):
Very nice refactor.

PR #372, opened 2021-06-01
[feat/aut964] Fixed set path in FilePicker

bertrandboudaud:
Changing path has no effect since we are assigning a temporary string.


bertrandboudaud:
@ahmedskhalil i know you have recently changed that maybe to fix one issue, please can you check if it breaks something

ghost (approved):
glad this was spotted sooner than later

PR #373, opened 2021-06-02
Fix/aut954 fixed crash when loading multiple sequence

bertrandboudaud:
Explorer Widget now uses the observer to refresh the data, in this way the search information is well synchronized with the session handler data.

another addition to fixing the crash was to move the session_handler_.setMinimalDataAndFilters before the event_dispatcher.dispatchEvents() so that the SessionHandler and the widget draw are synchronized as well.


ghost (approved):
problem mitigated

PR #374, opened 2021-06-07
[Feat/aut935] - Change testing framework to googletest

ghost:
Moving away from boost tests to googletest.


src/tests/class_tests/smartpeak/source/ParametersParser_test.cpp
```diff
@@ -17,92 +17,88 @@
...
+  EXPECT_STREQ(function_parameters[1].getDescription().c_str(), "a description");
+
+  EXPECT_STREQ(function_parameters[2].getName().c_str(), "param3");
+  EXPECT_STREQ(function_parameters[2].getType().c_str(), "float");
+  EXPECT_STREQ(function_parameters[2].getValueAsString().c_str(), "-1");
+  EXPECT_STREQ(function_parameters[2].getDescription().c_str(), "a quoted description, containing a comma");
 
   const string func2 {"func2"};
 
-  BOOST_REQUIRE(parameters.count(func2));
```
bertrandboudaud, line 64: This is quite minor, but BOOST_REQUIRE should be replaced by some ASSERT_* . The aim of using BOOST_REQUIRE is to stop the test as we will get into severe trouble and the rest of the test does not make sense (null pointer, out of range..) which ASSERT_ seems to be the counter part in google test. by the way, BOOST_REQUIRE was not so used in SmartPeak but I think it's a good practice since it add some readability to the test execution, which I think should not be lost, if possible.

bertrandboudaud (approved):
It's a nice move. the library looks better than boost test (container comparators ...).
please check my comment about the BOOST_REQUIREs. if you can do something.

src/tests/class_tests/smartpeak/source/ParametersParser_test.cpp
ghost: i'm aware of that

ChristopherAbram (approved):
Much more cleaner and readable tests! 👍

bertrandboudaud (approved):

bertrandboudaud:
That's great!

dmccloskey (approved):
Looks great.  Very nice change from Boost to GTest.

PR #376, opened 2021-06-15
Feat/aut966 Added support of NA keyword in the sequence file.

bertrandboudaud:
Added support of NA keyword in the sequence file.
* not case sensitive
* same effect as empty cells

**Note:** I had hard time with the CI; I needed to run it maybe 10 times to pass.
Either one test on ubuntu (LCMS_MRM_Unknown_test) or on macos (SequenceProcessor_test) or both were failing. it does not seem to be related to my minor modification... but if you see something suspicious in my changes ...


dmccloskey (approved):
Looks fine to me

PR #377, opened 2021-06-17
[Fix/aut963] - removing trailing zeros in FilePicker

ghost:
**Objective**
- Removing trailing zeros in FilePicker
- Passing unit test


src/smartpeak/source/core/Utilities.cpp
```diff
@@ -748,27 +748,37 @@ namespace SmartPeak
       if (entry_size >= 0 && entry_size < 1e3)
       {
         size_human_readable_stream << std::fixed << std::setprecision(2) << entry_size;
```
bertrandboudaud: the root cause is there, does it make any sense to have std::setprecision(2) for bytes (a file with 3.14 bytes cannot exist). its also nice to remove ".00" for other values however (MB, KB,...).

bertrandboudaud (commented):
good for me.

src/smartpeak/source/core/Utilities.cpp
ghost: it can but the system report file sizes in bytes so setting the precision is not needed since no conversion has taken place.

bertrandboudaud (approved):

PR #379, opened 2021-06-24
[fix/aut981] Fixed display of native_id for scans in explorer view

bertrandboudaud:
The case for the spectrum was missing. i replace that part of code with a generic computation of the table_entries_plot_col_ .


dmccloskey (approved):
At least that was a simple fix...

PR #385, opened 2021-07-03
[Feat/aut914] - Improving Feature Table

ghost:
**Objectives**
- Mapping Features Table Checkboxes to the Features Table.
- Reducing Feature Table's columns to `sample_name`, `sampe_type`, `component_group_name`, `replicate_group_name` and `component_name` as the default initial view.
- Samples shown in the table are reduced to 100 to save up on CPU cycles required to draw table entries.
- FilePicker's resiliency to invalid directory paths.

**Example View**

image


src/smartpeak/source/ui/Widget.cpp
```diff
@@ -144,6 +144,12 @@ namespace SmartPeak
     table_scanned_ = (table_data_.body_.dimension(0) == table_entries_.size() && !data_changed_);
 
     if (table_data_.body_.dimensions().TotalSize() > 0) {
+      if (100 < table_data_.body_.dimension(0)) {
```
bertrandboudaud: Would be nice to have a named constant (here or at the beginning of the file) that you can reuse here instead of repeating 3 times "100".

src/smartpeak/source/ui/Widget.cpp
```diff
@@ -144,6 +144,12 @@ namespace SmartPeak
...
+        print_until_ = 100;
+        ImGui::Text("Showing first 100 Entries For Performance");
```
bertrandboudaud: Would it be possible to display as well the total number of features?

src/examples/source/SmartPeakGUI.cpp
```diff
@@ -704,7 +704,10 @@ int main(int argc, char** argv)
     // feature table
     if (features_table_main_window_->visible_)
     {
-      exceeding_table_size_ = !session_handler_.setFeatureTable(application_handler_.sequenceHandler_, features_table_main_window_->table_data_);
+      bool data_changed = false;
+      exceeding_table_size_ = !session_handler_.setFeatureTable(application_handler_.sequenceHandler_,
```
bertrandboudaud, line 725: There was no way to use the observer pattern already used other places? and then to do the updateDataModificationState in the widget itself (we want to remove code as much as possible from GUI) It looks good this way, it's just not so consistent with the latest modifications.

bertrandboudaud (changes requested):
Looks Nice 👍 , the Feature Table seems to be more usable. also nice to have handled the performance issues. We should reuse your approach for other windows (transitions, injections ??...) when this issue can happen.
I had a few comments please check. the one about updateDataModificationState may require a lot of work. see if it's possible to have something more consistent. We should really decrease the code in the GUI, widget update being encapsulated, if possible.

src/smartpeak/include/SmartPeak/core/SessionHandler.h
```diff
@@ -109,11 +117,13 @@ namespace SmartPeak
...
 
-    @param[in] sequence_handler
+     @param[in] sequence_handler An instance of the SequenceHandler
+     @param[in] generic_table_data Current table data to be displayed
+     @param[out] data_changed true if the table has been altered
 
     @returns true if all rows/columns were added and false if rows/columns were omitted due to performance
     */
-    bool setFeatureTable(const SequenceHandler& sequence_handler, GenericTableData& generic_table_data);
+    bool setFeatureTable(const SequenceHandler& sequence_handler, GenericTableData& generic_table_data, EventDispatcher& event_dispatcher);
```
bertrandboudaud: Instead of having the Eventdispatcher here (which is a direct dependency to the GUI), you can use the IFeaturesObserver interface.

src/smartpeak/source/core/SessionHandler.cpp
```diff
@@ -1518,26 +1527,33 @@ namespace SmartPeak
...
+            LOGI << "Maximum permitted number of columns reached.";
             break;
           }
           for (int col = 0; col < n_cols; ++col) {
             if (row == 0) table_data.headers_(col) = headers.at(col);
             table_data.body_(row, col) = table.at(row).at(col);
           }
         }
+        event_dispatcher.onFeaturesUpdated();
+        this->onFeaturesUpdated();
```
bertrandboudaud: Is it necessary to have this this->onFeaturesUpdated();? (as well as this->onFeaturesSynced() below.) it will set the flag feature_table_updated  to true, but i don't see where feature_table_updated  is used (so this flag may be removed as well)

src/smartpeak/source/ui/Widget.cpp
```diff
@@ -144,6 +144,12 @@ namespace SmartPeak
     table_scanned_ = (table_data_.body_.dimension(0) == table_entries_.size() && !data_changed_);
 
     if (table_data_.body_.dimensions().TotalSize() > 0) {
+      if (max_rows_ < table_data_.body_.dimension(0)) {
+        print_until_ = max_rows_;
+        ImGui::Text("Showing first 100 samples for performance, total nr. of samples : %li", table_data_.body_.dimension(0));
```
bertrandboudaud: is it possible to also do a %li for the 100 here in this text?

src/smartpeak/include/SmartPeak/core/sources.cmake
```diff
@@ -9,6 +9,7 @@ set(sources_list_h
 	CastValue.h
 	ConsoleHandler.h
 	EventDispatcher.h
+  FeaturesObservable.h
```
bertrandboudaud: super minor: it loolk we have spaces here where tabs are used in the other lines.

src/smartpeak/include/SmartPeak/ui/Widget.h
```diff
@@ -200,6 +204,26 @@ namespace SmartPeak
...
+      if (observable) observable->addFeaturesObserver(this);
+    };
+
+    /**
+      IFeaturesObserver
+    */
+    virtual void onFeaturesUpdated() override
+    {
+      table_data_.clear();
+      data_changed_ = false;
```
bertrandboudaud, line 223: not sure to follow why we have to set data_changed_ to false here.

src/smartpeak/source/core/SessionHandler.cpp
```diff
@@ -1474,17 +1501,18 @@ namespace SmartPeak
...
+
   bool SessionHandler::setFeatureTable(const SequenceHandler & sequence_handler, GenericTableData& table_data)
   {
-    int MAX_SIZE = 5000;
     bool within_max_size = true;
     if (sequence_handler.getSequence().size() > 0 &&
       sequence_handler.getSequence().at(0).getRawData().getFeatureMapHistory().size() > 0) {
       // Make the feature table headers and body
-      if (feature_table_unique_samples_transitions_ != getNSelectedSampleNamesPlot()*getNSelectedTransitionsPlot() || table_data.body_.dimension(1) != 22 + getNSelectedFeatureMetaValuesPlot()) {
+      if (feature_table_unique_samples_transitions_ != getNSelectedSampleNamesPlot()*getNSelectedTransitionsPlot()
```
bertrandboudaud, line 1511: This test is not very accurate, we could profit from a flag set on the onFeaturesUpdated and replace this condition with it.

bertrandboudaud (approved):
I had 2-3 minor comments. would be nice if you can handle them (if it's easy), otherwise it could be done in a second time.

bertrandboudaud:
I really like the way you have limited the number of lines in the table. Sometimes, we can see some data with a lot of transitions for example that slows down the interface a lot. As SmartPeak is targeted to handle this kind of big data, it would be very nice to generalize that works to other (almost all) tables.

src/smartpeak/include/SmartPeak/ui/Widget.h
ghost: it doesn't cause the table to be re-updated constantly

src/smartpeak/include/SmartPeak/core/sources.cmake
ghost: i couldn't spot the space with the editor I am using so I changed its location

PR #388, opened 2021-07-06
[fix/aut982] Display error if required column is missing in the parameters file

bertrandboudaud:
properly handle error. the error is now displayed in the information and the log shows what column is missing


dmccloskey (approved):
Nice fix.

PR #389, opened 2021-07-06
Fixed show-back of log and info panel, once previously hidden

bertrandboudaud:


src/tests/class_tests/smartpeak/source/WindowSizesAndPositions_test.cpp
```diff
@@ -109,6 +109,22 @@ TEST(WindowSizesAndPositions, setTopWindowYSize)
   EXPECT_NEAR(win_size_and_pos.right_window_x_pos_, 100, 1e-3);
 }
 
+TEST(WindowSizesAndPositions, showHide)
```
dmccloskey, line 112: Nice addition

src/smartpeak/source/ui/WindowSizesAndPositions.cpp
```diff
@@ -58,6 +58,13 @@ namespace SmartPeak
     right_window_x_pos_ = left_window_x_size_ + bottom_and_top_window_x_size_;
   }
   void WindowSizesAndPositions::setWindowSizesAndPositions(const bool& show_top_window, const bool& show_bottom_window, const bool& show_left_window, const bool& show_right_window) {
+    // reset perc for bottom if needed
+    if (!show_bottom_window_ && show_bottom_window)
```
dmccloskey, line 62: Is the need for `show_bottom_window_` and the extra logic a bug on our side or a bug on the IMGUI side?

dmccloskey (approved):
Looks good from my end.  I had a small question I hope you don't mind answering :)

src/smartpeak/source/ui/WindowSizesAndPositions.cpp
bertrandboudaud: It's a bug on our side. once the bottom window is hidden, the top window takes all the space (this is our logic) and modifies the perc (space percentage that takes the bottom window), And then when we want to display the bottom window again, the perc stays to 0. with that modification, we can detect that we want to display back the bottom window and reset the perc.

bertrandboudaud: Another approach to have split windows would be to use that trick here: https://github.com/ocornut/imgui/issues/319. I think we could remove all these calculations we make by hand. It's not huge work, but maybe not so worth it as we have something working.

ghost (approved):

src/smartpeak/source/ui/WindowSizesAndPositions.cpp
dmccloskey: That looks like a promising approach.  If we can reduce code at the same time I think it would be well worth it.

PR #394, opened 2021-07-13
Feat/aut968 Automatic, accurate selection of data based on dilution and compound

bertrandboudaud:
This is implemented using a sampleGroupProcessor. indeed, we need to have all injections been proceeded to be able to select the features from other injections, so in the end, the RawDataProcessor was not a good candidate.

The user has to create a csv file named "selectiveDilutions.csv" containing the 2 columns ``compound`` and ``dilution_factor`` and select what compound is preferred on which dilution:

```
compound,dilution_factor
5-HTP,10
5-HTP2,10
...
```

then the features are selected: 

- if the compound is listed with the selected dilution, we will keep that one and ignore the other ones.
- if the compound is listed but the dilution is not listed, and if no other compound corresponds to the dilution, we will keep it.
- if the compound is not listed, we will keep it.

Note: while reviewing how mergeFeatures works, I have also renamed some variables, types, and reworked the layout of the code to make it more readable.


src/smartpeak/include/SmartPeak/core/Filenames.h
```diff
@@ -54,6 +54,7 @@ namespace SmartPeak
     std::string quantitationMethods_csv_i;
     std::string standardsConcentrations_csv_i;
     std::string referenceData_csv_i;
+    std::string selectiveDilutions_csv_i;
```
dmccloskey: I would recommend renaming to `selectDilutions`

src/smartpeak/include/SmartPeak/core/SampleGroupProcessor.h
```diff
@@ -75,29 +93,38 @@ namespace SmartPeak
       const Filenames& filenames
     ) const override;
 
+  private:
+
+    typedef std::tuple<std::string, std::pair<float, float>, float> mergeKeyType;
```
dmccloskey, line 98: Good idea

src/smartpeak/include/SmartPeak/io/SelectiveDilutionsParser.h
```diff
@@ -0,0 +1,52 @@
...
+// --------------------------------------------------------------------------
+
+#pragma once
+
+#include <map>
+#include <string>
+
+namespace SmartPeak
+{
+  class SelectiveDilutionsParser
```
dmccloskey: I would recommend renaming to `SelectDilutions` to match the filename or `SelectFromDilutions` to match the method name.  I would also recommend that we should just match the filename and method names to keep things simple once you finalize the name.

src/smartpeak/include/SmartPeak/pipelines/LCMS_MRM_Unknown_example.h
```diff
@@ -59,6 +59,7 @@ void example_LCMS_MRM_Unknowns(
   for (const InjectionHandler& injection : sequenceHandler.getSequence()) {
     const std::string& key = injection.getMetaData().getInjectionName();
     dynamic_filenames[key] = Filenames::getDefaultDynamicFilenames(
+      dir_I,
```
dmccloskey, line 62: I am assuming that this is for the users SmartPeak directory where all of their input files are located?

src/smartpeak/source/core/SampleGroupProcessor.cpp
```diff
@@ -26,10 +26,92 @@
...
+            const auto preferred_dilution = selective_dilution_map.at(native_id);
+            if (injection.getMetaData().dilution_factor == preferred_dilution)
+            {
+              add_feature = true;
+            }
+            else
+            {
+              // if there is no other feature is listed with the preferred dilution, we will use that one
+              const auto& existing_dilutions = existing_dilutions_map[native_id];
+              add_feature = (std::find(existing_dilutions.begin(), existing_dilutions.end(), preferred_dilution) == existing_dilutions.end());
```
dmccloskey: In most cases this should be a fair assumption.  However, I can also see this potentially leading to problems.  I would recommend adding in a parameter so that the user can control if they want to default to the available dilution if the preferred dilution is not present or simply omit the value.

src/smartpeak/source/io/SelectiveDilutionsParser.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <SmartPeak/io/SelectiveDilutionsParser.h>
+#ifndef CSV_IO_NO_THREAD
+#define CSV_IO_NO_THREAD
+#endif
+#include <SmartPeak/io/csv.h>
+#include <plog/Log.h>
+
+namespace SmartPeak
+{
+  const std::string s_compound{ "compound" };
```
dmccloskey: I would recommend renaming `compound` to `component_name` to be consistent with the other .csv files.  Also, we want to distinguish between a) Feature, PeptideRef, component_group_name, and transition_group_name, and b) subordinate, nativeID, component_name, and transition_name.  

As we you can see, we already have too many names for the same thing :)

dmccloskey (changes requested):
The logic and implementation of the "Select from dilutions" looks sound and well implemented.  I agree with your choice for adding it to the `SampleGroupProcessor` and also as a new method instead of modifying the `MergeInjections` method.

I had a few comments on allowing the user to modify the default behavior of the method and some minor naming suggestions.

Nice work.  Let me know if you have any questions on any of the comments.

src/smartpeak/include/SmartPeak/core/Filenames.h
bertrandboudaud: indeed.

src/smartpeak/include/SmartPeak/io/SelectiveDilutionsParser.h
bertrandboudaud: That's right.

src/smartpeak/include/SmartPeak/pipelines/LCMS_MRM_Unknown_example.h
bertrandboudaud: Yes. I need to be able to read that single file ("static") from the sampleGroupProcessor, which was never needed before in any other processor. So I had to introduce that directory. i was not sure to it the right way, but I haven't found any other example.

While doing this, I was thinking the Filenames class should be little bit reworked. I feel it's uncomfortable to have to declare the default file names in that Filename class when you add one processor. the processor should be a consistent, self declarative unit ("declarative" is important there: the Filename class can still have some responsibilities). in a way the processor should declare to the Filename class the file that it will need. and the Filename would construct the full path and check existence etc.. something like this. a JIRA task has been created for that we can discuss further.

src/smartpeak/source/io/SelectiveDilutionsParser.cpp
bertrandboudaud: :)
so i will rename to component_name

src/smartpeak/source/core/SampleGroupProcessor.cpp
bertrandboudaud: Added one method parameter to choose the behaviour
"preferred" the original behaviour 
"exlclusive" if the component is not found at the right dilution don't add it whatever dilution it is.

dmccloskey (approved):

PR #397, opened 2021-07-27
[Fix/aut979] - chromatogram's time-ranger in compact form

ghost:
**Objective**
Modifying chromatogram's time-ranger to a compact form.

**UI**
image


dmccloskey (approved):
Nice

PR #399, opened 2021-08-02
[fix/aut995] Fixed flooding of log when workflow fails to be loaded

bertrandboudaud:
log flooding is fixed.
also, the panel shows that an error occurred.

included in this PR:

- Workflow widget renamed WorkflowWodget for consistency
- Added unit test. I don't like testing the protected methods and even more protected member variables, but we have no choice here (for UI objects). I think it makes sense if we select the ones that are very key to the internal logic (here: error flag is checked).

image


dmccloskey (approved):
In my opinion, the strategy of using protected methods instead of private methods to enable the use of unit testing is a suitable compromise.  Looks good.

PR #401, opened 2021-08-02
[Feat/aut531] - Multithreading options for SequenceProcessor methods

ghost:
**Objectives**

- Parser to deduce the number of threads from the `parameters` object of the first injection in the sequence segment or sample group.
- Implement a multi-thread launcher similar to that used by `ProcessSequence` for both `ProcessSequenceSegment` and `ProcessSampleGroup` methods.
- Update the unit tests to test for the expected number of threads and workers.


dmccloskey (approved):
Nice to have full multi-threading capabilities now.

PR #403, opened 2021-08-10
[feat/aut985] - Update Github Template with instructions to attach log files

ghost:
**Objectives**
- Users now aren't required to compress log files since they are generated as log files.


bertrandboudaud (approved):
Nice

PR #423, opened 2021-09-21
Feat/aut156 - Implementation of User Session

### LINKED ISSUE #425: [SmartPeakCLI] - Fails to load featureQCComponents.csv when featureQCComponentGroups.csv is not provided.
**Describe the bug**
When only featureQCComponents.csv is provided to SmartPeakCLI on Linux, SmartPeak will fail to perform QC on the data. Only when featureQCComponents.csv is provided AND featureQCComponentGroups.csv is provided will SmartPeak load the QC file and check the QC rules.

This only manifests on Linux.

**To Reproduce**
1. Go to the LCMS_MRM_Unknowns SmartPeak example dataset
2. Rename / delete featureQCComponentsGroups.csv to something that SmartPeak will not recognize, but keep featureQCComponents.csv in the directory as-is
3. On Linux, use SmartPeakCLI to load sequence.csv and run the workflow.
4. Open resulting PivotTable.csv OR FeatureDB.csv and examine the QC_transition_pass and QC_transition_score values
5. Notice that all the values for these two columns are 1, indicating that the QC system was not invoked
6. Repeat steps on Windows, but this time there are various values in QC_transition_pass and QC_transition_score columns, indicating that QC works.

**Expected behavior**
Loads featureQCComponents.csv and computes QC_transition_pass and QC_transition_score without issue.

**logs**
Attach all recent SmartPeak log files which can be found in the following locations :
 - Windows: [Windows_smartpeak_log_2021-09-22_16-17-20.csv](https://github.com/AutoFlowResearch/SmartPeak/files/7214368/Windows_smartpeak_log_2021-09-22_16-17-20.csv)
 - Linux: [Linux_smartpeak_log_2021-09-22_15-37-07.csv](https://github.com/AutoFlowResearch/SmartPeak/files/7214343/Linux_smartpeak_log_2021-09-22_15-37-07.csv)

**FeatureDB.csv data results**
- [LCMS_MRM_Unknowns_FeatureDB_linux.csv](https://github.com/AutoFlowResearch/SmartPeak/files/7214346/LCMS_MRM_Unknowns_FeatureDB_linux.csv)
- [LCMS_MRM_Unknowns_FeatureDB_windows.csv](https://github.com/AutoFlowResearch/SmartPeak/files/7214347/LCMS_MRM_Unknowns_FeatureDB_windows.csv)

**Version information**
 - OS: CentOS Linux 7 (Core) and Windows 10
 - SmartPeak Version 1.12.0

bertrandboudaud:
The user now has to create a session to use SmartPeakGUI.

1. The user creates a New session with File -> New Session 
2. The User chooses the location
image
3. The session file widget is displayed using the file that is in the directory, the user can then choose to change the value, to embed or not the data in the database (by default, it's not embedded).
image
4. Some checks are made on the file existence to avoid a wrong setup. The user can only press ok when all is ready.
image
5. The session is created, but the user has to save it with File -> save as
image

When loading a session, if a file is missing, the session files widget appears showing the missing file. the user can fix it and save it.

A Save status is displayed on the title of the window.
At the moment, the status changes when you edit the parameters or the workflow. since no other table can be edited, these are the only ones.
image

When running a workflow, the sessions files are checked and if one file is missing to run the workflow, a popup appears.
image

On the CLI side, nothing changed except that load-session now uses the session database file instead of the sequence file. It's also possible to pass a directory, where the file is organized as it was previously.


src/smartpeak/include/SmartPeak/core/ApplicationProcessor.h
```diff
@@ -41,14 +40,14 @@ namespace SmartPeak
...
     virtual ~ApplicationProcessor() = default;
 
-    // Each of the derived classes implement the following virtual methods
     virtual bool process() = 0;
 
     /* IProcessorDescription */
-    int getID() const override { return -1; }
-    std::string getDescription() const override { return ""; }
-    ParameterSet getParameterSchema() const override { return ParameterSet(); };
+    virtual std::string getDescription() const override { return ""; }
```
dmccloskey, line 46: Nice updates

src/smartpeak/include/SmartPeak/core/WorkflowManager.h
```diff
@@ -82,6 +82,16 @@ namespace SmartPeak {
     */
     void updateApplicationHandler(ApplicationHandler& source_app_handler);
 
+    /**
+      @brief Returns the list of requirements to run the workflow.
+    */
+    std::set<std::string> getRequirements(const std::vector<ApplicationHandler::Command>& commands) const;
+
+    /**
+      @brief return true if the Filenames contains all the requirements to reun the commands
```
dmccloskey:
```suggestion
      @brief return true if the Filenames contains all the requirements to run the commands
```

src/smartpeak/include/SmartPeak/io/SessionDB.h
```diff
@@ -0,0 +1,314 @@
...
+    void logSQLError(const std::string& error_message, const std::string& sql_command = "") const;
+
+  protected:
+    std::filesystem::path session_file_name_;
+    std::string smartpeak_version_ = "Unknown";
+    bool session_info_logged_ = false;
+  };
+
+  template<typename Value, typename ...Args>
+  std::optional<SessionDB::DBContext> SessionDB::beginWrite(const std::string& table_name, const Value& value, const char* value_type, const Args& ...args)
```
dmccloskey, line 127: Thanks for isolating the methods below the class definition.

src/smartpeak/include/SmartPeak/ui/SessionFilesWidget.h
```diff
@@ -0,0 +1,106 @@
...
+      std::string text_editor_last_update_; // used to detect a change in the textfield (cb does not work properly)
+      std::string original_filepath_;
+      bool embedded_ = false;
+      bool original_embedded_ = false;
+      bool file_exists_ = false;
+      std::string combo_file_mode_ = "";
+    };
+
+    Filenames filenames_;
+    std::optional<std::string> popup_file_picker_;
```
dmccloskey, line 64: I have noticed that `std::optional` is used quite a bit.  Is it required due to the SQL light integration code?

src/smartpeak/source/cli/Task.cpp
```diff
@@ -33,361 +33,374 @@
...
-            {
-                handle_integrity_check(
-                    run_integrity_check(
-                        application_handler.sequenceHandler_, 
-                        SmartPeak::InputDataValidation::heavyComponentsAreConsistent,
-                        "Run IS integrity check"),
-                    application_settings.allow_inconsistent,
-                    "IS integrity check failed!");
-            }
+          if (all_integrity_check || ApplicationSettings::contains_option(integrity_args, "SAMPLE", int_msg))
```
dmccloskey, line 125: I missed this previously.  What are your thoughts on using ENUM here instead of a std::string for the second argument?  This is very small so please ignore if you think it is not important.

src/smartpeak/source/core/ApplicationHandler.cpp
```diff
@@ -0,0 +1,196 @@
...
+        break;
+      }
+    }
+    return saved;
+  }
+
+
+  void ApplicationHandler::setFileSavedState(const std::string& file_id, bool saved_state)
+  {
+    saved_files_[file_id] = saved_state;
```
dmccloskey: I would recommend using `emplace` here instead of indexing.

src/smartpeak/source/core/FeatureFiltersProcessorUtils.cpp
```diff
@@ -0,0 +1,329 @@
...
+        else
+        {
+          featureQCFile.store(filenames.getFullPath(file_id).generic_string(), features_qc, false);
+        }
+      }
+      if (!filenames.getFullPath(file_group_id).empty())
+      {
+        if (filenames.isEmbedded(file_group_id))
+        {
+          auto db_context = filenames.getSessionDB().beginWrite(
```
dmccloskey: Is this logic able to handle "on the fly" user defined filters?

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -396,6 +397,11 @@ namespace SmartPeak
     LOGD << "END LoadFeatures";
   }
 
+  std::vector<std::string> StoreFeatures::getRequirements() const
```
dmccloskey, line 400: This is a nice addition.  I can also see how this could be used to define a much more interactive workflow builder at some point in the future.

dmccloskey (commented):
I have taken a first pass through the code.  It looks nice and I can see it was a lot of work!  I left some comments.  I am sure I missed several things so hopefully they do not appear too random.  I still need to review the tests and perhaps a second round if that makes sense.

src/smartpeak/include/SmartPeak/ui/SessionFilesWidget.h
bertrandboudaud: No it's not due to the SQL light integration code, I just find it useful. std::optional is used quite a lot when something may not be available, or when a method can return an object or nothing (in case of error).
in that case, a flag was needed to say I want to open the filepicker with that file_id (ImGui cannot open correctly the filepicker when it's needed so we need a variable to postpone it). if the option is not set, setting the option on will open the filepicker with the right file_id. if the option is not set, the file picker will not open. So in the end, it avoids the declaration of 2 variables for one action.

src/smartpeak/source/cli/Task.cpp
bertrandboudaud: Yes, we can use Enum there.
but I would prefer to create a ticket for this (I Will create).

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: Yes, actually, all the processor are filled with the same requirement, almost, "sequence" and "traML".

src/smartpeak/source/core/FeatureFiltersProcessorUtils.cpp
bertrandboudaud: Do you mean the user can add more "properties" (columns)? if yes, then this is an issue. if you mean that we can add more filters and use this method, yes.

dmccloskey: The user can add more "properties" using the "metavalue_" expression.  There is a section in the tabular view of the UI where a dictionary is iterated over with the different "metavalue_" attributes that have been defined by the user.  This is supported in the reader/writer for the OpenMS:FeatureFilter class.

ghost (approved):
Great improvements and nice to see session import/export in place, I imagine the `session_info` table will in future implementations include more information about the session state such as currently opened plots/tables or was it meant to be for other uses?

src/smartpeak/source/core/FeatureFiltersProcessorUtils.cpp
bertrandboudaud: OK, so I will implement it (hopefully, one day should be enough) before merging it. I think I will implement it in a separate table, as these are "dynamic".

bertrandboudaud:
> Great improvements and nice to see session import/export in place, I imagine the `session_info` table will in future implementations include more information about the session state such as currently opened plots/tables or was it meant to be for other uses?

Yes, it can be used for that :)

PR #427, opened 2021-09-29
[Feature request] for SmartPeakGUI - "select all" button when generating Report

### LINKED ISSUE #419: [Feature request] for SmartPeakGUI - "select all" button when generating Report
Would be nice to have a single button to select all Sample Type and Metadata to be included in the Report generated from the GUI.

ghost:
**Objective**
Select-/Unselect-All for sample type and metadata to be exported through the report widget.

**Linked Issue**
[[Feature request] for SmartPeakGUI - "select all" button when generating Report #419](https://github.com/AutoFlowResearch/SmartPeak/issues/419)

image

### DIFF
--- a/src/smartpeak/source/core/Utilities.cpp
+++ b/src/smartpeak/source/core/Utilities.cpp
@@ -862,25 +862,30 @@ namespace SmartPeak
     return filenames;
   }
 
+  std::string Utilities::makeUniqueStringFromTime()
+  {
+    const std::time_t time_now = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+    auto time_in_epochs = std::chrono::system_clock::now().time_since_epoch();
+    char time_str[32];
+    strftime(time_str, 32, "%Y-%m-%d_%H-%M-%S_", std::localtime(&time_now));
+    return std::string(time_str + std::to_string(time_in_epochs.count()));
+  }
+
   std::filesystem::path Utilities::createEmptyTempDirectory()
   {
-    std::filesystem::path path;
-    uint i;
-    uint max_tries = 1000;
-    while (true)
+    std::filesystem::path tmp_dir_path;
+    auto path_to_tmp_dir = std::filesystem::path(std::string("tmp_").append(makeUniqueStringFromTime()));
+    tmp_dir_path = std::filesystem::temp_directory_path();
+    std::filesystem::current_path(tmp_dir_path);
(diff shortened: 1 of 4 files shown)


src/smartpeak/source/core/Utilities.cpp
```diff
@@ -862,25 +862,30 @@ namespace SmartPeak
...
+    const std::time_t time_now = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+    auto time_in_epochs = std::chrono::system_clock::now().time_since_epoch();
+    char time_str[32];
+    strftime(time_str, 32, "%Y-%m-%d_%H-%M-%S_", std::localtime(&time_now));
+    return std::string(time_str + std::to_string(time_in_epochs.count()));
+  }
+
   std::filesystem::path Utilities::createEmptyTempDirectory()
   {
-    std::filesystem::path path;
```
bertrandboudaud, line 867: It's nice you had improved this :)

one question is about the way you construct the temporary directory here. you are using a string from time, but it seems to be a second precision string, is it ? if yes it, it can be an issue as you may construct 2 temporary directories within the same second (with unit test it can happen easily).

bertrandboudaud (approved):
Looks good to me.

Please see however my comments about the empty temporary directory and the 2 below, in case you feel easy to improve it.

About the UI, I think the checkbox is not well adapted for this action. A checkbox reflects a status on/off, while here it's used an action. however it works and it's already the way it is used in other places, so I guess it's ok.

Would be nice also if the check box would be in a separate space - outside the scrolling area - (or maybe just a separator is missing) to avoid mixing action with the properties.

src/smartpeak/source/core/Utilities.cpp
ghost: Therefore the time since epoch is appended to create a unique string for the temp directory.

ghost:
> Looks good to me.
> 
> Please see however my comments about the empty temporary directory and the 2 below, in case you feel easy to improve it.
> 
> About the UI, I think the checkbox is not well adapted for this action. A checkbox reflects a status on/off, while here it's used an action. however it works and it's already the way it is used in other places, so I guess it's ok.
> 
> Would be nice also if the check box would be in a separate space - outside the scrolling area - (or maybe just a separator is missing) to avoid mixing action with the properties.

I'd say it's more natural to have the All checker a part of the rest, the user would go up to click on All or scroll down to choose individual elements.

PR #433, opened 2021-10-11
[Feat/aut1022] - SmartPeak support for exporting workflow results as plots - Part II

ghost:
**Objective**
- Exporting heatmap plots

**SmartPeak Heatmap Widget View**
image

**Exported File**
smartpeak-exported-plot

### DIFF
--- a/src/smartpeak/include/SmartPeak/ui/Plotter.h
+++ b/src/smartpeak/include/SmartPeak/ui/Plotter.h
@@ -203,6 +203,8 @@ namespace SmartPeak {
     std::string y_title_;
     std::string plot_title_;
     std::string gnuplot_path_;
+    uint        file_width_;
+    uint        file_height_;
     PlotType    plot_type_;
     
     std::vector<PlotLineProperties> plotlines_properties_;
--- a/src/smartpeak/include/SmartPeak/ui/Plotter.h
+++ b/src/smartpeak/include/SmartPeak/ui/Plotter.h
@@ -109,8 +109,7 @@ namespace SmartPeak {
     
     enum class PlotType {
       CURVE,
-      HEATMAP,
-      HISTOGRAM
+      HEATMAP
     };
   
     PlotExporter(std::string output_path, SessionHandler::GraphVizData& graphvis_data, int format, PlotType plot_type);
--- a/src/smartpeak/include/SmartPeak/ui/Plotter.h
+++ b/src/smartpeak/include/SmartPeak/ui/Plotter.h
@@ -106,7 +106,16 @@ namespace SmartPeak {
   class PlotExporter
   {
   public:
-    PlotExporter(std::string output_path, SessionHandler::GraphVizData& graphvis_data, int mode);
+    
+    enum class PlotType {
+      CURVE,
+      HEATMAP,
+      HISTOGRAM
+    };
+  
+    PlotExporter(std::string output_path, SessionHandler::GraphVizData& graphvis_data, int format, PlotType plot_type);
+    
+    PlotExporter(std::string output_path, SessionHandler::HeatMapData& graphvis_data, int format, PlotType plot_type);
     
     /*
     @brief starts plotting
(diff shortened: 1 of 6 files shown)


src/smartpeak/include/SmartPeak/ui/Plotter.h
```diff
@@ -106,7 +106,16 @@ namespace SmartPeak {
   class PlotExporter
   {
   public:
-    PlotExporter(std::string output_path, SessionHandler::GraphVizData& graphvis_data, int mode);
+    
+    enum class PlotType {
+      CURVE,
```
dmccloskey, line 111: What is "CURVE"?  Is this for the line plots or scatter plots?  If so, I would recommend naming them something like "Line" or "Scatter" or even better would be to use the same as what ImPlot uses.

src/smartpeak/source/ui/Plotter.cpp
```diff
@@ -295,7 +338,7 @@ namespace SmartPeak {
     if (fout.is_open()) {
       appendFileHeader_(fout);
       if (isTermAvailable_("cairo")) {
-        fout << "set terminal pngcairo dashed enhanced" << std::endl;
+        fout << "set terminal pngcairo size 1400,800 dashed enhanced" << std::endl;
```
dmccloskey: I am not familiar with the GNU API, so please pardon the question if it does not make sense.

Is `fout` storing the png parameters in string format that are then parse by GNU?  I am thinking it might be nice as an intermediate step to set the hard coded values for size, font type, etc for future work to allow for customizing the plots by the users.

dmccloskey (commented):
It looks nice so far.  I had just a few minor comments.  I will let @bertrandboudaud have the final review.

src/smartpeak/include/SmartPeak/ui/Plotter.h
ghost: By "CURVE" we mean straightened line plots as well as curves (parabolas, sin waves, etc..) since both fall into the same category. I would prefer to keep it since changing it to "LINE" can be misleading but let me know if we should give it a better descriptive name.

src/smartpeak/source/ui/Plotter.cpp
ghost: That is correct, what we do here is to generate a gnuplot script that gnuplot can interpret. The reason why the size of the exported media has been set this way is because gnuplot does't seem to care whether the titles and legends are off-margin or within the visible range. I can change this to member variables that can be set during object initialisation.

src/smartpeak/include/SmartPeak/ui/Plotter.h
```diff
@@ -106,7 +106,16 @@ namespace SmartPeak {
...
+      HEATMAP,
+      HISTOGRAM
```
bertrandboudaud: is HISTOGRAM used? if not, would be better to remove it (or handle it as an error when used, which is not the best solution).

src/smartpeak/include/SmartPeak/ui/Plotter.h
bertrandboudaud: It's nice to have same wording for these plots and our internal wording. in the session handler it's Plot, ScatterPlot or Heatmap. We can live with the CURVE or LINE as well.

bertrandboudaud (changes requested):
It Looks nice for me.

Please see my comment about HISTOGRAM, which I think needs some action (the reason of Request Changes).

the discussion about LINE/CURVE, I'm not sure how can find a good solution, even if I feel wee need to be consistent as much as possible across our interfaces, one side we use plot on the other we use CURVE.

src/smartpeak/include/SmartPeak/ui/Plotter.h
ghost: It's not used yet so I will remove it.

bertrandboudaud (approved):

PR #434, opened 2021-10-11
[fix/aut1031] Fixed computation of merge injections

bertrandboudaud:
Fixed computation of merge injections when scan_mass high and low are set to -1. (to my understanding, it happens when the parameters of the injection match a "wildcard" injections key that we construct at the begining. 

Also fixed 2nd pass merging that was using scan_polarity_merge_feature_name parameter instead of mass_range_merge_feature_name


ghost (approved):

dmccloskey (approved):
Thanks again for looking into this.  Glad to see it appears to be a special case.

PR #435, opened 2021-10-13
Fix dda parameters

bertrandboudaud:
Usage of parameters was missing in the DDA processor.

Also, some special handling has been added to support relative path for the following parameters:
AccurateMassSearchEngine db:mapping
AccurateMassSearchEngine db:struct
AccurateMassSearchEngine positive_adducts
AccurateMassSearchEngine negative_adducts

in this way, when moving the directory, there is no need to edit the parameters (and therefore, the FIAMS and DDA example work out of the box)

DDA example comes with a simplified DB that only handle Germicidin A.

Germicidin A is well identified.
image


src/smartpeak/include/SmartPeak/core/Utilities.h
```diff
@@ -400,5 +400,24 @@ namespace SmartPeak
      @brief return an empty, temporary directory.
     */
     static std::filesystem::path createEmptyTempDirectory();
+
+    /**
+      @brief for parameter that is supposed to be file path, make it full path if it exists.
+    */
+    static void prepareFileParameter(
```
dmccloskey, line 407: Nice additions

src/tests/class_tests/smartpeak/source/Utilities_test.cpp
```diff
@@ -643,3 +643,82 @@ TEST(utilities, removeTrailing)
   Utilities::removeTrailing(str3, ".00");
   EXPECT_STREQ(str3.c_str(), "234");
 }
+
+TEST(utilities, prepareFileParameter)
```
dmccloskey, line 647: Nice

dmccloskey (approved):
Nice bug fix and nice additions with the relative path.

PR #437, opened 2021-10-14
[Feat/aut951] - Distributed Mode

ghost:
**Objective**
- SmartPeakServer App
- Offload computational workloads to a dedicated entity running on HPCs as a microservice.
- Modifications to SmartPeakGUI with enabled communications to the SmartPeakServer via HTTP2 to ensure lightweight operation on the client side.


src/examples/CMakeLists.txt
```diff
@@ -1,8 +1,29 @@
...
+#
+# --------------------------------------------------------------------------
+# $Maintainer: Ahmed Khalil, Bertrand Boudaud $
+# $Authors: Douglas McCloskey $
+# --------------------------------------------------------------------------
 
 cmake_minimum_required(VERSION 3.8.2 FATAL_ERROR)
 project("SmartPeak_class_examples_smartpeak")
-
+#cmake_policy(SET CMP0115 OLD)
```
bertrandboudaud: is it still needed? if yes, can you add some comment on that

src/examples/CMakeLists.txt
```diff
@@ -119,19 +192,22 @@ add_dependencies(INTERACTIVE_EXAMPLES ${interactive_executables_list})
...
-      # set_target_properties(${_examples} PROPERTIES MACOSX_BUNDLE_INFO_PLIST ${SMARTPEAK_HOST_DIRECTORY}/cmake/MacOSX/Info.plist)
+            PRIVATE ImPlot::Sources ImGui::Sources ImGui::Examples ImGui::SourcesMiscCpp
+            ${grpc_required_libs} OpenSSL::SSL OpenSSL::Crypto)
       install(TARGETS ${_examples}  BUNDLE DESTINATION .  COMPONENT Applications)
     else()
       add_executable(${_examples} source/${_examples})
-      target_link_libraries(${_examples} PUBLIC ${SmartPeak_LIBRARIES} OpenMS)
+      target_link_libraries(${_examples} PUBLIC ${SmartPeak_LIBRARIES} OpenMS 
+						PRIVATE 
+						${grpc_required_libs})
```
bertrandboudaud: do the examples need grpc dependency?

src/examples/source/SmartPeakGUI.cpp
```diff
@@ -694,7 +700,77 @@ int main(int argc, char** argv)
       
       ImGui::EndMainMenuBar();
     }
+      
+    // ======================================
+    // Server-Side Execution
+    // ======================================
+      if (run_workflow_widget_->server_fields_set)
```
bertrandboudaud, line 716: would it be possible to put that block of code in a dedicated (several?) class?

src/examples/source/service/services.hpp
```diff
@@ -0,0 +1,490 @@
+// --------------------------------------------------------------------------
```
bertrandboudaud, line 1: Can we have a header and cpp file?

src/smartpeak/source/core/Server.cpp
```diff
@@ -0,0 +1,305 @@
...
+// $Authors: Ahmed Khalil $
+// --------------------------------------------------------------------------
+
+#include <SmartPeak/core/Utilities.h>
+#include <SmartPeak/core/Server.h>
+
+namespace SmartPeak {
+  namespace serv {
+  
+    bool contains_option(
```
bertrandboudaud: the 3 methods contains_option, extract_report_sampletypes and extract_report_metadata should follow the global coding standard that is start with non capital leter, then no underscore but capital letter as a separator

protos/workflow.proto
```diff
@@ -0,0 +1,74 @@
+
+syntax = "proto3";
+package SmartPeakServer;
+
+service Workflow {
+  rpc runWorkflow (WorkflowParameters) returns (WorkflowResult) {}
+  rpc getLogStream (InquireLogs) returns (stream LogStream) {}
+}
+
```
bertrandboudaud, line 38: do we need all this below? report, graph data .. all this should be local there should not be any message about this, as i understood.

src/examples/source/service/workflow.grpc.pb.cc
```diff
@@ -0,0 +1,247 @@
+// Generated by the gRPC C++ plugin.
```
bertrandboudaud, line 1: As these files are generated, is it worth putting them in the repository? at first, I would say no, but I don't have a strict opinion on this. But if we put them in our repository, how is it generated? if it's generated by hand, would be nice to document in an appropriate place (maybe inline documentation in the .proto file would be a convenient location)

bertrandboudaud (changes requested):
Niced work, please see my comments.

src/examples/CMakeLists.txt
ghost: No, it was a reminder to append source code suffixes to file name, recent versions of CMake (>3.19) will produce lots of warnings when that's not the case which is annoying when debugging.

src/examples/CMakeLists.txt
ghost: Yes they do.

src/examples/source/SmartPeakGUI.cpp
ghost: I have packed some of the lines into separate functions, it's not possible to do that for the whole block since we don't want to have the gRPC library as part of the libSmartPeak or libSmartPeakWidgets.

src/examples/source/service/services.hpp
ghost: This file binds directly to the SmartPeakServer executable, making a library out of it is not necessary and it's not a good idea to have it as part of the main SmartPeak library.

protos/workflow.proto
ghost: They will be needed later when the other approaches for transmitting the visualisation data that we have discussed is implemented.

src/examples/source/service/workflow.grpc.pb.cc
ghost: A short notice on how to generate them is now included in the proto file.

ghost: These files are need for the SmartPeakServer and SmartPeakGUI.

protos/workflow.proto
dmccloskey: Can you elaborate on the need for this here?

src/examples/CMakeLists.txt
```diff
@@ -105,6 +127,55 @@ find_package(ImPlot COMPONENTS Sources REQUIRED)
 include_directories(${IMPLOT_INCLUDE_DIR})
 
 #------------------------------------------------------------------------------
+# gRPC
```
dmccloskey, line 130: Can these sections be moved to their own CMake file?

src/examples/CMakeLists.txt
dmccloskey: This concerns me a bit.  Can you elaborate on why the examples would need GRPC dependencies?  From my understanding, none of the examples use GRPC?

src/examples/source/service/services.hpp
dmccloskey: I would also recommend a header/cpp file.  If for some reason that is not possible, I would recommend using the same convention but placing what would be the .cpp contents after the .h contents.

dmccloskey (commented):
The updates look really nice.  In addition, looks like the unit tests for the new Utility methods are missing?  If so can you add them as well.

ghost:
> The updates look really nice. In addition, looks like the unit tests for the new Utility methods are missing? If so can you add them as well.

Thanks for reminding, I have added the unittests for the new utility methods.

dmccloskey (commented):
Thanks for adding in the unit tests.  This is looking good from my end.  Let's wait on Bertrands feedback and then go from there.

bertrandboudaud (approved):
ok for me

PR #439, opened 2021-10-15
Feat/aut997 Display Errors in Info Panel

bertrandboudaud:
Errors are now displayed on the info panel.
details like injection, sample group ... and processor name are displayed along with the error message.

image

However - not all:
We are catching exceptions from the processor and displaying the message of it. the user at least is aware that something has been stopped, so the rule for a processor is when an error is fatal to its execution, a std::exception is thrown. Errors that are not fatal (which kind of - should be avoided) will not be displayed in the info panel.

Some workarounds have been made to support this also when loading sessions (as it does not use SequenceProcessor where the error handling is made). It should be acceptable since, at some point, we may want to use the SequenceProcessor also to load the sessions (and profit from the progress bar...).

Along with this modification, START and END logs in the processors have been removed for several reasons:
- The mechanism of exception requires properly log the end of the processor to keep the log readable.
- at the same time, more context is added to the START and END logs.
- less code, less chance of making errors, and reduced maintenance.

### DIFF
--- a/src/smartpeak/source/core/RawDataProcessor.cpp
+++ b/src/smartpeak/source/core/RawDataProcessor.cpp
@@ -2548,7 +2548,7 @@ namespace SmartPeak
       f.setSubordinates(f_map.second);
       fmap.push_back(f);
     }
-
+    rawDataHandler_IO.setFeatureMap(fmap);
     rawDataHandler_IO.updateFeatureMapHistory();
 
     LOGI << "MergeFeatures output size: " << rawDataHandler_IO.getFeatureMap().size();
--- a/src/smartpeak/source/core/RawDataProcessor.cpp
+++ b/src/smartpeak/source/core/RawDataProcessor.cpp
@@ -334,8 +334,7 @@ namespace SmartPeak
 
     if (!InputDataValidation::prepareToStore(filenames_I, "mzML_i"))
     {
-      LOGD << "END " << getName();
-      return;
+      throw std::invalid_argument("Failed to store output file");
     }
 
     OpenMS::MzMLFile mzmlfile;
@@ -366,8 +365,7 @@ namespace SmartPeak
 
     if (!InputDataValidation::prepareToLoad(filenames_I, "featureXML_i"))
     {
-      LOGD << "END " << getName();
-      return;
+      throw std::invalid_argument("Failed to load input file");
     }
 
     try {
@@ -410,8 +408,7 @@ namespace SmartPeak
 
     if (!InputDataValidation::prepareToStore(filenames_I, "featureXML_o"))
     {
-      LOGD << "END " << getName();
-      return;
+      throw std::invalid_argument("Failed to store output file");
     }
 
     // Store outfile as featureXML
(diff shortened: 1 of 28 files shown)


src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -652,37 +625,30 @@ namespace SmartPeak
...
-        std::vector<OpenMS::MRMFeatureSelector::SelectorParameters> p =
-          Utilities::extractSelectorParameters(params_I.at("MRMFeatureSelector.schedule_MRMFeatures_score"), params_I.at("MRMFeatureSelector.select_MRMFeatures_score"));
-        OpenMS::MRMBatchFeatureSelector::batchMRMFeaturesScore(rawDataHandler_IO.getFeatureMap(), output, p);
-      }
-      else 
-      {
-        LOGE << "Both arguments 'select params' and 'schedule params' are empty";
-        LOGD << "END selectFeatures";
-        // TODO: replace throw with return?
-        throw std::invalid_argument("Both arguments 'select params' and 'schedule params' are empty.");
```
dmccloskey, line 681: Throwing an exception is fine.  We just want to make sure that all of the errors from each of the processors (or from OpenMS) are caught so that the application does not crash.

src/smartpeak/source/core/RawDataProcessor.cpp
```diff
@@ -2495,58 +2415,53 @@ namespace SmartPeak
...
-            s.setMetaValue("mz_error_ppm", hit.getMetaValue("mz_error_ppm"));
-            s.setMetaValue("mz_error_Da", hit.getMetaValue("mz_error_Da"));
-            s.setCharge(hit.getCharge());
-            std::vector<OpenMS::Feature> subs = { s };
-            f.setSubordinates(subs);
-            fmap.push_back(f);
+    // Run the accurate mass search engine
+    OpenMS::MzTab output;
+    ams.init();
+    ams.run(rawDataHandler_IO.getFeatureMap(), output);
```
dmccloskey, line 2421: How are exceptions from OpenMS handled now that the try/catch block is removed?

dmccloskey (commented):
I like the direction.  I had a few comments in regards to error handling particularly with OpenMS that I would like to understand a bit better.

src/smartpeak/source/core/RawDataProcessor.cpp
bertrandboudaud: When an exception occurs, it is now caught by the caller (processSegment, processSampleGroups, processInjection), as std::exceptions.
OpenMS exceptions all derive from std::exception so we can catch them all along with their message.

On the processSegment, processSampleGroups, processInjection side, the exception is caught, then error observer notified, and then they will continue. So the behavior stays the same, it shouldn't crash much as previously.

One case that does not work properly is if openms logs an error but does not throw any exception, In that case the error will still be logged as previously, but will not appear in the info panel.

Or if one openms method is supposed to be returning false on error and we do not handled that properly on our side (we should throw an exception in that case) which will ends with the same behavior: error in the log, but not in the panel - this case we can fix it on our side.

dmccloskey: Thanks so much for that explanation.

dmccloskey (approved):

PR #446, opened 2021-10-19
Fix/select dilution

bertrandboudaud:
The way to select features depending on dilution preference has been reworked and is now integrated with the merge injections.
2 new parameters have been added: ``select_preferred_dilution`` (false by default, true to use it), and the ``select_preferred_dilutions_file``.

The ``select_preferred_dilutions_file`` holds the list of components and the dilution that we want to prefer to select.
image

It's now possible to export a Feature table and Pivot Table for Groups. (GUI only)

Note: exporting group pivot table or feature table is not supported for CLI.

### Example of use
This is the sequence file given as an example:
 
image

The parameters are set as follow:
image

**Note:** the 2 new parameters: ``select_preferred_dilution``, if you want to apply the dilution rule, to set to true. And the ``select_preferred_dilutions_file`` that sets the file you want to apply (relative to the working directory or full path). 

**Note** set all merge rules to MAX. in fact it is needed if you have more than one injection with the preferred dilution (like in this example), then it will select the max one.
 
This is the dilution file:
 
image

This is the Workflow :
image

**Note:** the select dilution step does not exist anymore, the select dilution is now part of the merge_injections step. it will then create a feature group.
 
To produce the result: use the new buttons Group Feature DB or Group Pivot Table:
 
image

And this is the result (GroupPivotTable.csv):
 
image

As a reminder:
 
image

If we run with this dilution file:
 
image

This is the difference:
(left with dilution_factor to 10, right with the dilution_factor to 1)
 
image

We can see the max of dilution 1 has been selected.


dmccloskey (changes requested):
Very nice work re-working this method.  In my opinion, the way in which you integrated the `selectDilutions` part does a nice job of taking advantage of the `MergeInjections` method so that it can be stacked with other merge logic.

This method is not the most intuitive.  Would you mind adding some lines to the Class/method description (perhaps even copy and paste from your PR text) describing how to use the method and configure the parameters to get the desired behavior for the proper selection of dilutions?

Other than that it looks great.

dmccloskey (approved):
Thanks for the updated the docs

PR #450, opened 2021-10-26
Fix/fix sample group processing

bertrandboudaud:
Running FIAMS Workflow on MacOS leads to an infinite loop on merging injections.

It seems to be due to the SampleGroupProcessorMultithread called from inside a sample group iteration loop.
another side effect about that was the progress bar jumping from different, not monotonic, increasing values.

In this PR, is also fixed:
- a crash if you expend details of the progress information while the first event is not yet received (at the beginning of the workflow)
- ProcessSequenceSegments and ProcessSampleGroups use spawn_workers instead of run_processor which makes it use threading processing.

### DIFF
--- a/src/smartpeak/source/core/SequenceProcessor.cpp
+++ b/src/smartpeak/source/core/SequenceProcessor.cpp
@@ -41,7 +41,7 @@
 
 namespace SmartPeak
 {
-
+  
   class WorkflowException : public std::exception
   {
   public:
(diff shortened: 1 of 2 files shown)


src/smartpeak/source/core/SequenceProcessor.cpp
```diff
@@ -106,34 +129,14 @@ namespace SmartPeak
...
-    }
-
     // Launch
+    notifySequenceProcessorStart(injections.size());
     SequenceProcessorMultithread manager(
       injections,
       filenames_,
       raw_data_processing_methods_,
       this);
+    int n_threads = nbThreads(injections);
```
dmccloskey, line 139: +1 for the modularization

dmccloskey (approved):
Good bug fixes.

PR #452, opened 2021-11-08
Feat/aut1004 - Save layout in session

bertrandboudaud:
An interface has been introduced, ``IPropertiesHandler``, and each class inheriting from this can describe the properties that can be serialized in the session DB.

The session DB can both read and write ``IPropertiesHandler``.

The ``SessionLoaderGenerator`` is a class that will construct the loading workflow to be able to load the computed features when restoring the session.

``SetInputOutputWidget`` is a widget that allows the user to change the mzml, feature in, and feature output directories. this widget is used as another step in the ``LoadSessionWizard``. very similar to the ``RunWorkflowWidget`` we could at some point factorize it (the role though is quite different).

``ExplorerWidget`` has been moved in a separated, dedicated file.

``SplitWindow`` is a class that holds the ``WindowSizesAndPos`` class as well as the list of top, bottom, left, and right windows.


src/smartpeak/include/SmartPeak/iface/IPropertiesHandler.h
```diff
@@ -0,0 +1,41 @@
...
+#pragma once
+
+#include <SmartPeak/core/CastValue.h>
+
+#include <map>
+#include <optional>
+
+namespace SmartPeak
+{
+  struct IPropertiesHandler
```
dmccloskey, line 36: Would it be possible to add a description for the struct (if applicable as many of the other interfaces do not) as well as method docstrings?

src/smartpeak/include/SmartPeak/io/SessionDB.h
```diff
@@ -59,6 +59,9 @@ namespace SmartPeak
       return session_file_name_;
     }
 
+    bool writePropertiesHandler(const IPropertiesHandler& properties_handler);
```
dmccloskey, line 66: Would it be possible to add docstrings to these new methods?

src/smartpeak/include/SmartPeak/ui/ExplorerWidget.h
```diff
@@ -0,0 +1,106 @@
...
+*/
+
+namespace SmartPeak
+{
+  extern bool enable_quick_help;
+
+  /**
+    @brief Base class for all tables
+
+    TODO: features
```
dmccloskey: Are these my notes?  I think they can be removed.

dmccloskey: Also it looks like the class description is not in accordance with the actual method.

src/smartpeak/include/SmartPeak/ui/LayoutLoader.h
```diff
@@ -0,0 +1,127 @@
...
+  struct LayoutLoader : IApplicationProcessorObserver
+  {
+    LayoutLoader(ApplicationHandler& application_handler) :
+      application_handler_(application_handler)
+    { };
+
+    /**
+    IApplicationProcessorObserver
+    */
+    virtual void onApplicationProcessorStart(const std::vector<std::string>& commands) override
```
dmccloskey: Is there a reason the methods are defined in the header file?

src/smartpeak/include/SmartPeak/ui/SetInputOutputWidget.h
```diff
@@ -0,0 +1,71 @@
...
+#include <SmartPeak/core/ApplicationHandler.h>
+#include <SmartPeak/ui/Widget.h>
+#include <SmartPeak/ui/FilePicker.h>
+
+#include <string>
+#include <vector>
+
+namespace SmartPeak
+{
+  struct ISetInputOutputWidgetObserver
```
dmccloskey, line 37: Would it be possible to add some additional docstrings for the classes and methods here?

src/smartpeak/include/SmartPeak/ui/SplitWindow.h
```diff
@@ -0,0 +1,52 @@
...
+#include <SmartPeak/ui/Widget.h>
+#include <SmartPeak/ui/LayoutLoader.h>
+#include <SmartPeak/ui/WindowSizesAndPositions.h>
+
+#include <string>
+#include <vector>
+
+namespace SmartPeak
+{
+  class SplitWindow final : public Widget
```
dmccloskey, line 37: This solution is nicer than what we had previously in my opinion.  If it would be possible to add some class documentation that would be great.

src/smartpeak/source/core/ApplicationProcessor.cpp
```diff
@@ -389,6 +379,74 @@ namespace SmartPeak
...
+        p.second.setTagValue(Filenames::Tag::FEATURES_INPUT_PATH, application_handler_.filenames_.getTagValue(Filenames::Tag::FEATURES_INPUT_PATH));
+        p.second.setTagValue(Filenames::Tag::FEATURES_OUTPUT_PATH, application_handler_.filenames_.getTagValue(Filenames::Tag::FEATURES_OUTPUT_PATH));
+      }
+    }
+    std::set<std::string> injection_names;
+    for (const auto& injection : application_handler_.sequenceHandler_.getSequence())
+    {
+      injection_names.insert(injection.getMetaData().getInjectionName());
+    }
+    const std::set<std::string> sequence_segment_names = {}; // session_handler_.getSelectSequenceSegmentNamesWorkflow(application_handler_.sequenceHandler_);
```
dmccloskey: Are the commented lines needed?

src/smartpeak/source/core/SessionLoaderGenerator.cpp
```diff
@@ -0,0 +1,100 @@
...
+        loading_workflow_commands_.emplace(20, "MAP_CHROMATOGRAMS");
+      }
+      else if (command == "EXTRACT_CHROMATOGRAM_WINDOWS")
+      {
+        loading_workflow_commands_.emplace(21, "EXTRACT_CHROMATOGRAM_WINDOWS");
+      }
+      else if (command == "ZERO_CHROMATOGRAM_BASELINE")
+      {
+        loading_workflow_commands_.emplace(22, "ZERO_CHROMATOGRAM_BASELINE");
+      }
```
dmccloskey, line 57: "MERGE_SPECTRA" and "EXTRACT_SPECTRA_WINDOWS" from the FIA-MS workflow could also be added.  Their inverse commands are the same.

src/smartpeak/source/io/SessionDB.cpp
```diff
@@ -193,25 +193,292 @@ namespace SmartPeak
...
+  auto properties = properties_handler.getPropertiesSchema();
+  os.str("");
+  os << "CREATE TABLE ";
+  os << "\"" << table_name << "\"";
+  os << " (";
+  os << "ID INTEGER PRIMARY KEY, ";
+  std::string separator;
+  for (const auto& [column_name, column_type] : properties)
+  {
+    // beginWrite(os, db_context.columns, value, value_type, args...);
```
dmccloskey: Is this comment needed?

src/smartpeak/source/io/SessionDB.cpp
```diff
@@ -193,25 +193,292 @@ namespace SmartPeak
...
+    sqlite3_finalize(db_context.stmt);
+    db_context.stmt = nullptr;
   }
+  closeSessionDB(db_context.db);
 
-  int64_t SessionDB::getLastInsertedRowId(SessionDB::DBContext& db_context) const
+  return true;
+}
+
+bool SessionDB::readPropertiesHandler(IPropertiesHandler& properties_handler)
```
dmccloskey, line 369: The read/writePropertiesHandler introduces a lot of new code to the class.  Are you sure it is not appropriate to reuse the code you had already written for reading and writing the the SQLite DB?

src/smartpeak/source/ui/LoadSessionWizard.cpp
```diff
@@ -29,43 +29,95 @@ namespace SmartPeak
...
+        load_session.notifyApplicationProcessorError("Failed to load session.");
       }
-      else
+      load_session.notifyApplicationProcessorEnd();
+    }
+  }
+
+  void LoadSessionWizard::onInputOutputCancel()
+  {
+
```
dmccloskey: Is this intended?

dmccloskey (changes requested):
From the review, it looks like that was quite a bit of work!  Very nice!

In addition to the comments that I left, it looks like several of the new classes/methods do not have matching unit tests.  If unit tests are appropriate, it would be great to include them in this PR.  We can always hold off on additional sprint items if this takes more time than expected.

src/smartpeak/include/SmartPeak/iface/IPropertiesHandler.h
bertrandboudaud: sure

src/smartpeak/include/SmartPeak/ui/LayoutLoader.h
bertrandboudaud: Fixed

src/smartpeak/include/SmartPeak/ui/SplitWindow.h
bertrandboudaud: Yes, I was thinking first I had to serialize it, but in the end, no. but I kept it will be useful later.

src/smartpeak/source/io/SessionDB.cpp
bertrandboudaud: the other set of API is actually not so easy to reuse as it is templated, unfortunately. I would suggest one day unifying this with one set of API (the new one is better)

src/smartpeak/source/ui/LoadSessionWizard.cpp
bertrandboudaud: Yes, actually the LoadSessionWizard will just not continue, so it stays empty (it's not possible, one day we had to clean something). I'm not for doing a function that does nothing, but as its method is an observer method interface, it has to be overridden. and to remove it from the observer API may be also confusing.

bertrandboudaud:
> From the review, it looks like that was quite a bit of work! Very nice!
> 
> In addition to the comments that I left, it looks like several of the new classes/methods do not have matching unit tests. If unit tests are appropriate, it would be great to include them in this PR. We can always hold off on additional sprint items if this takes more time than expected.

I had fixed all your comment please see the last commits.
Yes maybe some methods are missing tests. However I made a verification that most of them and the most important are covered. I would go like this and have a second round If you agree.

dmccloskey (approved):
Lets go for a second round then 👍

PR #455, opened 2021-11-11
Enable OpenMP for MacOS Builds

### LINKED ISSUE #405: "FIAMS Unknowns" workflow using SmartPeak on a set of mzML files
Hi,

I'm trying to run the "FIAMS Unknowns" workflow using SmartPeak v1.11.0 on a Mac with macOS Big Sur Version 11.5.1. I have a set of mzMl files and when I want to run the workflow the app crashes. I have attached my most recent log file. I'd appreciate any help on that :)
[smartpeak_log_2021-08-11_14-52-33.csv](https://github.com/AutoFlowResearch/SmartPeak/files/6968626/smartpeak_log_2021-08-11_14-52-33.csv)

### LINKED ISSUE #453: SmartPeak GUI Crashing on MacOS
**Describe the bug**
I'm running the SmartPeak GUI on an Intel Macbook Pro 16" MacOS Big Sur 11.6.1. When I try to run a workflow on the sample data, the program immediately closes (crashes?). No output files are created. This happens as soon as a I click "Run Workflow" on the popup window asking for the input mzML, input features, and output features folders 

**To Reproduce**
1. Open SmartPeak GUI
2. File -> New Session
3. Select session directory from Sample Data and click "Ok" (have tried "LCMS_MRM_QCs", "LCMS_MRM_Standards", "LCMS_MRM_Unknowns", and "HPLC_UV_Standards"). 
4. Actions -> Run Workflow
5. Confirm mzML folder, Input features folder, and Output features folder click "Run workflow"
6. Program crashes (happening every time)

**Expected behavior**
I expect the workflow to run at this point and to create output files/results

**logs**
SmartPeak crash log
[smartpeak_log_2021-11-08_15-27-04.log](https://github.com/AutoFlowResearch/SmartPeak/files/7503435/smartpeak_log_2021-11-08_15-27-04.log)

Apple Crash Report Log
[smartpeak_apple_crash_log_211109.log](https://github.com/AutoFlowResearch/SmartPeak/files/7503437/smartpeak_apple_crash_log_211109.log)

**data**
I am using the SmartPeak sample data for now

**Version information**
 - OS: MacOS 11.6.1 (Macbook Pro 16" Intel i7)
- SmartPeak - 1.16.0

bertrandboudaud:
Unlike other platforms, on MacOS OpenMP was not enabled for OpenMS, as we see in this build log: 

```
-- OpenMP support requested: ON
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_CXX_FOUND CXX) 
```

OpenMP seems to be a hard requirement as it is used, for example in their logging system to be thread-safe.

At the moment, users report crashes on MacOS, and call stacks show most of the time a crash around openMS's logger methods. we can also see sometimes interlaced log messages from openMS, just before crashing, which looks worrying.

As in this log:
```
While loading '/Users/distiller/SmartPeak/src/examples/data/LCMS_MRM_QCs/mzML/150601_0_BloodProject01_PLT_QC_Broth-1.mzML': Ill formed absolute or relative sourceFile path: file://.le://.
```
(note the double "le://." at the end, coming from a previous log message).

So it is a strong indication that at least one of the reasons for these crashes is the missing OpenMP when building OpenMS. 

one side note: only to install the openmp developer lib is sufficient to include it in the build. that would explain why it works on some machines (developer machines?) and not on others. anyway, as race conditions are involved, machine model, os version, all can lead to different behaviors.

This PR adds a stress test that indeed crashes only on Mac when OpenMP is not there, using the example data. so I had to mix example data with unit tests, which may be a little bit weird depending on the plan we had about the different data usages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,7 +109,7 @@ jobs:
             brew remove yarn node
             brew install cmake qt5 sdl2
             brew install cppcheck gnuplot
-            pip3 install -r tools/smartpeak/requirements.txt
+            brew install libomp            pip3 install -r tools/smartpeak/requirements.txt
       - run:
           name: "Pull latest (only release)"
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,7 +109,6 @@ jobs:
             brew remove yarn node
             brew install cmake qt5 sdl2
             brew install cppcheck gnuplot
-            brew install libomp
             pip3 install -r tools/smartpeak/requirements.txt
       - run:
           name: "Pull latest (only release)"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,6 +109,7 @@ jobs:
             brew remove yarn node
             brew install cmake qt5 sdl2
             brew install cppcheck gnuplot
+            brew install libomp
             pip3 install -r tools/smartpeak/requirements.txt
       - run:
           name: "Pull latest (only release)"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -132,7 +132,8 @@ jobs:
             tar -xzf contrib_build.tar.gz
             sed -i '' 's/std::auto_ptr/std::unique_ptr/g' include/seqan/basic/test_system.h
             cd ~/OpenMS &&  mkdir openms_debug_build && cd openms_debug_build &&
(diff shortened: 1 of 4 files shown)


dmccloskey (commented):
Would it be possible to simply run the examples similar to Ubuntu for MacOS instead of adding in a new unit test?

bertrandboudaud:
> Would it be possible to simply run the examples similar to Ubuntu for MacOS instead of adding in a new unit test?

Yes, I think it's a better idea. i will make the change.

ghost (changes requested):
Let's remove the line `sed -i '' 's/#pragma omp parallel for//g' src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp` in `config.yml` to reinstate full OpenMP support.

bertrandboudaud:
> Let's remove the line `sed -i '' 's/#pragma omp parallel for//g' src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp` in `config.yml` to reinstate full OpenMP support.

Ok, done

bertrandboudaud:
> Would it be possible to simply run the examples similar to Ubuntu for MacOS instead of adding in a new unit test?

It's done now. And indeed the examples fail if we don't activate OpenMP. So I removed the unit test.

ghost (approved):

PR #456, opened 2021-11-15
Display error in info panel when user select incorrect session file

bertrandboudaud:
The original bug report was about a crash when the user selects a file that is not a session file (for example, he selects  sequence.csv file).

Now the crash seems to have disappeared but no error is displayed on the info panel. This pr fixes it.


dmccloskey (approved):
Nice work.  The additional test to cover the bug is much appreciated :)

PR #459, opened 2021-11-25
Moved processors into separate files.

bertrandboudaud:
Moved processors into separate files.

inside the ``core`` directory (both header and sources), 4 subdirectories have been created:
``ApplicationProcessors``
``RawDataProcessor``
``SampleGroupProcessors``
``SequenceSegmentProcessors``

Inside each of the directories, on file is created for each processor.

As it is a PR with lot of renaming, it's not easy to read, sorry!
But basically, nothing related with the code is involved, only some inclusion has been reworked/cleared.


dmccloskey (approved):
Nice work!  A much needed refactor given all of our work the past months.

PR #464, opened 2021-12-08
AUT-993 - UI Splitting and Docking

bertrandboudaud:
This new layout uses the Docking branch of imgui, which is WIP, but stable enough to be used as described here:
https://github.com/ocornut/imgui/wiki/Docking

User can:

- Resize panels (improved compared to the in-house made code)
- Dock tabs to other panel
- Dock out windows
- Split windows to more get more complex layout

Layout is saved in the session db, as previously, but using in file from imgui, which handle docking settings.

Window -> Reset Window Layout will reset the layout as it is as usual, with the 3 panels, left, top, bottom.

Docking

### DIFF
--- a/external/imgui.cmake
+++ b/external/imgui.cmake
@@ -3,13 +3,12 @@ include(${CMAKE_ROOT}/Modules/ExternalProject.cmake)
 set_property (DIRECTORY PROPERTY EP_BASE Dependencies)
 
 set (DEPENDENCIES)
-#set (IMGUI_VERSION 6a0d0dab5a9f0b9518a2bc9bb456a69895ae0962) # tag: v1.72b, 2019/07/31
-set (IMGUI_VERSION v1.81) # tables
+set (IMGUI_VERSION 848d21b6b56d369fb7976bd88b5f4094ab9ea50a) # on Docking branch
 message (STATUS "Adding imgui ${IMGUI_VERSION} as an external project.")
 
 ExternalProject_Add(imgui
   GIT_REPOSITORY https://github.com/ocornut/imgui.git
-  GIT_TAG 848d21b6b56d369fb7976bd88b5f4094ab9ea50a
+  GIT_TAG ${IMGUI_VERSION}
   UPDATE_COMMAND ""
   CONFIGURE_COMMAND ""
   BUILD_COMMAND ""
--- a/external/imgui.cmake
+++ b/external/imgui.cmake
@@ -9,7 +9,7 @@ message (STATUS "Adding imgui ${IMGUI_VERSION} as an external project.")
 
 ExternalProject_Add(imgui
   GIT_REPOSITORY https://github.com/ocornut/imgui.git
-  GIT_TAG ${IMGUI_VERSION}
+  GIT_TAG 848d21b6b56d369fb7976bd88b5f4094ab9ea50a
   UPDATE_COMMAND ""
   CONFIGURE_COMMAND ""
   BUILD_COMMAND ""
--- a/src/examples/source/SmartPeakGUI.cpp
+++ b/src/examples/source/SmartPeakGUI.cpp
@@ -178,8 +178,8 @@ int main(int argc, char** argv)
   auto spectrum_explorer_window_ = std::make_shared<ExplorerWidget>("SpectrumExplorerWindow", "Scans", &event_dispatcher);
   auto sequence_main_window_ = std::make_shared<SequenceTableWidget>("SequenceMainWindow", "Sequence",
                                                                       &session_handler_, &application_handler_.sequenceHandler_);
-  auto transitions_main_window_ = std::make_shared<GenericTableWidget>("TransitionsMainWindow", "Transitions");
-  auto spectrum_main_window_ = std::make_shared<GenericTableWidget>("SpectrumMainWindow", "Scans");
+  auto transitions_main_window_ = std::make_shared<GenericTableWidget>("TransitionsMainWindow", "Transitions Table");
+  auto spectrum_main_window_ = std::make_shared<GenericTableWidget>("SpectrumMainWindow", "Scans Table");
   auto quant_method_main_window_ = std::make_shared<SequenceSegmentWidget>("QuantMethodMainWindow", "Quantitation Method",
                                    &session_handler_, &application_handler_.sequenceHandler_,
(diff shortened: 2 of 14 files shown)


external/imgui.cmake
```diff
@@ -9,7 +9,7 @@ message (STATUS "Adding imgui ${IMGUI_VERSION} as an external project.")
 
 ExternalProject_Add(imgui
   GIT_REPOSITORY https://github.com/ocornut/imgui.git
-  GIT_TAG ${IMGUI_VERSION}
+  GIT_TAG 848d21b6b56d369fb7976bd88b5f4094ab9ea50a
```
ghost: Would it be possible to pull from a git tag instead? This has caused problems in the past as the whole commit history had to be pulled to find this exact one.

src/examples/source/SmartPeakGUI.cpp
```diff
@@ -178,8 +178,8 @@ int main(int argc, char** argv)
   auto spectrum_explorer_window_ = std::make_shared<ExplorerWidget>("SpectrumExplorerWindow", "Scans", &event_dispatcher);
   auto sequence_main_window_ = std::make_shared<SequenceTableWidget>("SequenceMainWindow", "Sequence",
                                                                       &session_handler_, &application_handler_.sequenceHandler_);
-  auto transitions_main_window_ = std::make_shared<GenericTableWidget>("TransitionsMainWindow", "Transitions");
-  auto spectrum_main_window_ = std::make_shared<GenericTableWidget>("SpectrumMainWindow", "Scans");
+  auto transitions_main_window_ = std::make_shared<GenericTableWidget>("TransitionsMainWindow", "Transitions Table");
```
ghost, line 181: A better idea is to substitute all the window/widget string IDs ("TransitionsMainWindow", "Transitions Table", ..) with enum classes to make sure they are unified across all other classes/functions.

src/examples/source/SmartPeakGUI.cpp
```diff
@@ -672,6 +672,15 @@ int main(int argc, char** argv)
       }
       showQuickHelpToolTip("actions");
       
+      if (ImGui::BeginMenu("Window"))
+      {
+        if (ImGui::MenuItem("Reset Window Layout"))
```
ghost, line 677: Nice.

src/widgets/source/ui/Heatmap2DWidget.cpp
```diff
@@ -146,15 +146,16 @@ namespace SmartPeak
...
-            ImPlot::ShowColormapScale(heatmap_data_.feat_value_min_, heatmap_data_.feat_value_max_, height_ - 70);
+            ImPlot::ShowColormapScale(heatmap_data_.feat_value_min_, heatmap_data_.feat_value_max_, window_size.y - 70);
             float available_width = ImGui::GetContentRegionAvailWidth();
             ImGui::SameLine();
             const ImPlotFlags imPlotFlags = ImPlotAxisFlags_LockMin | ImPlotAxisFlags_LockMax | ImPlotAxisFlags_TickLabels;
             if (ImPlot::BeginPlot(plot_title_.c_str(),
               heatmap_data_.feat_heatmap_x_axis_title.c_str(),
               heatmap_data_.feat_heatmap_y_axis_title.c_str(),
-              ImVec2(available_width - 80, height_ - 70),
+              ImVec2(available_width - 80, window_size.y - 70),
```
ghost, line 158: It would be nice to derive the constant values automatically whenever possible. If not it would be very useful to declare them with descriptive names or document what they are for.

ghost (commented):
Very nice work moving the windows and tables around.

external/imgui.cmake
bertrandboudaud: I don't think there is a tag there unfortunately.
by the way, I think I should set the IMGUI_VERSION to the commit hash instead of putting the hash here.

src/widgets/source/ui/Heatmap2DWidget.cpp
bertrandboudaud: I agree. However, i will create another task for that, as this is not something introduced in that PR and is out of the scope of this feature for now.

src/examples/source/SmartPeakGUI.cpp
bertrandboudaud: Indeed the thing here is because we have 2 windows with the same name (Scans/Scans), (Transitions/Transitions) then the docker mechanism was lost and displays these windows at the wrong place.
So we have to have a unique name across the application.
However, an enum looks to be a very static way to solve this issue. It's possible that one day we want to have multiple instances of the same type of window, very dynamical (could be "chromatogram 1", "chromatogram 2" ...) and using an enum may be a problem there. perhaps a warning could be done instead. with some check in the main gui.

src/widgets/source/ui/Heatmap2DWidget.cpp
bertrandboudaud: I agree. I have created a separate task for that on JIRA (as this PR didn't introduce that, and it would be out of scope to fix it now).

src/examples/source/SmartPeakGUI.cpp
dmccloskey: We would definitely want a dynamic solution to ensure unique names for each widget in the future once we introduce the ability to copy and split windows.

src/widgets/include/SmartPeak/ui/Widget.h
```diff
@@ -403,6 +399,7 @@ namespace SmartPeak
   */
   static void showQuickHelpToolTip(const std::string& ui_element_name)
   {
+    return;
```
dmccloskey: Is this correct?

dmccloskey (changes requested):
That is nice to see that ImGUI now takes care of all of the heavy lifting in regards to the UI window calculations, etc.  From my end, everything looks good.  There does appear to be an injected return statement in the `showQuickHelpToolTip` that looks to be a mistake (but I could be wrong).

src/widgets/include/SmartPeak/ui/Widget.h
bertrandboudaud: oh, good catch, I put it this because I was hunting a bug by reducing ImGui widgets.

dmccloskey (approved):

PR #466, opened 2021-12-12
Support for small screens

bertrandboudaud:
Starting size of the main window on 75% width and height of the screen
added exit menu item

before, no way to access window's title bar and resizing controls, no exit possibility
image

after
image


dmccloskey (approved):
Small change but well needed.

PR #467, opened 2021-12-14
Feat/aut1052 Enable Viewports

bertrandboudaud:
Thanks to imgui's Viewport system (on Docking branch), it's now possible to drop out windows from the main window.

window positioning is absolute instead of relative.
another nice side effect is that all the popups also are displayed out of the main window.

capt_chromatogram


dmccloskey (approved):
That is a great break for us!

PR #469, opened 2021-12-15
Fixed Cancel button size for directory selection dialog box

bertrandboudaud:
Little esthetic fix on the cancel button that is not fully displayed when we select a new session directory.

the incorrect size comes for 2 reasons:
- PushItemWidth does not work for buttons, you have to pass size in the Button() function
- The main field takes 90% of the width which let not enough space for the buttons anyway

before: (the cancel button is cropped)
image

fixed:
image


dmccloskey (approved):
Nice fix.

PR #470, opened 2021-12-15
Increased max points to be displayed to 10 millions

bertrandboudaud:
imgui is by default limited to 64k vertices display, which prevents us to display large plots (overflow displays a corrupted graph).

It's possible and safe to break this limitation by changing ImDrawIdx to unsigned int

```
 // Your renderer backend will need to support it (most example renderer backends support both 16/32-bit indices).
 // Another way to allow large meshes while keeping 16-bit indices is to handle ImDrawCmd::VtxOffset in your renderer.
 // Read about ImGuiBackendFlags_RendererHasVtxOffset for details.
//#define ImDrawIdx unsigned int
```

the number of points is increased to 10.000.000
tested with a graph using 2.500.000 points, and it looks smooth enough on a DTU machine.

### DIFF
--- a/external/imgui.cmake
+++ b/external/imgui.cmake
@@ -13,4 +13,5 @@ ExternalProject_Add(imgui
   CONFIGURE_COMMAND ""
   BUILD_COMMAND ""
   INSTALL_COMMAND ""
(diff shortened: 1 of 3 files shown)


dmccloskey (approved):
Another nice fix.

PR #475, opened 2021-12-28
[Feat/aut1007] - Marking highest values in 2D Plots

ghost:
**Objective**

- Marking and displaying the highest values in Spectras efficiently.

**Sample View using `FIAMS_FullScan_Unknowns`**
image
image


dmccloskey (commented):
Very nice.  The ability to indicatete the highest value for each sample/feature combo is nice.

Instead of (or in addition to depending on what is more aesthetically pleasing), can you add in the y-value for the highest peak?

ghost:
> Very nice. The ability to indicatete the highest value for each sample/feature combo is nice.
> 
> Instead of (or in addition to depending on what is more aesthetically pleasing), can you add in the y-value for the highest peak?

Thanks for your proposal, I have updated the plots to draw the highest y-values along the marking.

dmccloskey (approved):
Looks good.

PR #480, opened 2022-01-05
Report Doalog box now confirms success or error once the report is created

bertrandboudaud:
The report dialog box was missing some feedback.
no notification on success or on error. the user had to check the log (which most probably will not do).
it's problematic when a report is opened with Excel, you don't know that you failed to overwrite it.

image

One issue on that is the size_t that is been set to -1 on error in the cvswriter's writeDataInRow which is incorrect since it's an unsigned type. A std::optional has been used there.


dmccloskey (approved):
Nice changes.  The issue with Excel can be very annoying unless you already know what is going on.

PR #508, opened 2022-01-22
Feat/improving plot zoom

### LINKED ISSUE #506: Use ImPlot zooming feature
We want to use the facilities provided by ImPlot to navigate into the chromatograms and the spectrograms.

bertrandboudaud:
Using the implot zoom feature.

(also works for spectrograms)

ChromZom

### DIFF
--- a/src/widgets/source/ui/GraphicDataVizWidget.cpp
+++ b/src/widgets/source/ui/GraphicDataVizWidget.cpp
@@ -25,7 +25,6 @@
 #include <SmartPeak/ui/FilePicker.h>
 #include <imgui.h>
 #include <imgui_internal.h>
-#include <implot.h>
 
 namespace SmartPeak
 {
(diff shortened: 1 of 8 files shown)


src/widgets/source/ui/GraphicDataVizWidget.cpp
```diff
@@ -25,45 +25,29 @@
...
     updateData();
-    drawSliders();
+    drawGraphHeader();
     drawGraph();
   }
 
-  void GraphicDataVizWidget::drawSliders()
+  void GraphicDataVizWidget::drawGraphHeader()
   {
-    if (ImGui::DragFloatRange2("Time-Range",
```
dmccloskey, line 42: Great to see we no long have to track this type of logic.

dmccloskey (approved):
Very nice spot on this update from ImPlot.  That takes care of a lot of things we were having to handle ourselves in sometimes not the best way.

PR #512, opened 2022-02-01
Fix/check standard concentrations

### LINKED ISSUE #499: Providing wrong standardsConcentrations.csv does not throw any error
**Describe the bug**
If the user provide an incorrect standardsConcentrations.csv, running LCMS_MRM_Standards does not raise any error and the resulting output quantitationMethods.csv is the one provided as input.

**To Reproduce**
1. in the LCMS_MRM_Standards  example, replace the standardsConcentrations.csv with the following file
[incorrect_standardsConcentrations.csv](https://github.com/AutoFlowResearch/SmartPeak/files/7886997/incorrect_standardsConcentrations.csv). It contains incorrect headers.
2. With SmartPeak, open the LCMS_MRM_Standards example 
3. Execute the workflow
4. Check log and info panel. No error is displayed
5. Check the feature directory, the segment1_quantitationMethods.csv file is the one provided as an input.

**Expected behavior**
An error is displayed in both the log and the info panel segment1_quantitationMethods.csv may not be produced.

**Version information**
1.18

bertrandboudaud:
Some OpenMS loader methods do not raise any error when loading data. We add some tests on SmartPeak to check that the CSV files have the required headers.


src/smartpeak/include/SmartPeak/core/Utilities.h
```diff
@@ -475,5 +481,30 @@ namespace SmartPeak
...
     static bool isList(const std::string& str, const std::regex& re);
+
+    /**
+     @brief Check if a csv file has required headers
+    */
+    template<char Separator, typename ...Columns>
+    static bool checkCSVHeader(const std::filesystem::path& filename, const Columns& ...columns);
   };
+
+  template<char Separator, typename ...Columns>
```
dmccloskey, line 497: Nice use of templates

dmccloskey (approved):
Looks good.  It looks like there are a few conflicts that need to be fixed.  My apologies if I looked at the reviews in the wrong order :).  Please merge when the conflicts are resolved.

PR #513, opened 2022-02-01
Feat/check csv file

### LINKED ISSUE #499: Providing wrong standardsConcentrations.csv does not throw any error

### LINKED ISSUE #500: Providing UTF-8 with BOM csv does not raise any error.
**Describe the bug**
Some users provide sometimes UTF8, with BOM, csv files. in most cases, it's not handled and it results in some weird behavior.

**To Reproduce**
1. in the LCMS_MRM_Standards example, replace the file "standardsConcentrations.csv" with this one: 
[standardsConcentrations.csv](https://github.com/AutoFlowResearch/SmartPeak/files/7887056/standardsConcentrations.csv). It contains UTF8 BOM.
2. In SmartPeak open the LCMS_MRM_Standards  example
3. Run the Workflow
4. Check the Log; no error is displayed.
5. open segment1_quantitationMethods.csv in the feature folder, it's not updated (note: because it's in an error state, it shouldn't even be produced, see https://github.com/AutoFlowResearch/SmartPeak/issues/499)

**Expected behavior**
Support UTF8 with BOM or raise an error, as much as possible. It's possible that some parts of the code are not SmartPeak responsibility, but we have some CSVReaders that can check that. Also, we could check files that we provide to the OpenMS library.

**Version information**
1.18

Note obviously, in most of the cases, the 1st column of the CSV will be ignored, and most probably an error would be raised. This report describes one case where no error is displayed at all. It would be also convenient for the user to know why he gets this error because it's usually confusing since when you open the file with some text editor, you cannot see where is the problem.

bertrandboudaud:
It happened some times that user provides a file with the UTF8 BOM marker, which introduce a lot of confusion as the 1st column will not be read properly and it's not obviously visible when opening the file with a text editor.
We add some check and raise error if csv files have the BOM marker.


dmccloskey (approved):
Looks good.

PR #515, opened 2022-02-17
Feat/improve calibration2

bertrandboudaud:
- Ability manually to exclude points from calibration.
- Changed axes legend.
- Ability to edit parameters (global and quantmethods)
- Ability to show chromatogram from one point.

image

image

image


src/smartpeak/include/SmartPeak/core/SessionHandler.h
```diff
@@ -341,10 +341,27 @@ namespace SmartPeak
     */
     struct CalibrationData
     {
-      std::vector<std::vector<float>> conc_raw_data;
-      std::vector<std::vector<float>> feature_raw_data;
-      std::vector<std::vector<float>> outlier_conc_raw_data;
-      std::vector<std::vector<float>> outlier_feature_raw_data;
+
+      struct Points
```
dmccloskey, line 345: Nice idea with the nested struct for points.

src/smartpeak/source/core/SequenceSegmentProcessors/CalculateCalibration.cpp
```diff
@@ -45,7 +45,20 @@ namespace SmartPeak
   ParameterSet CalculateCalibration::getParameterSchema() const
   {
     OpenMS::AbsoluteQuantitation oms_params;
-    return ParameterSet({ oms_params });
+    ParameterSet parameters({ oms_params });
```
dmccloskey: I would recommend not modifying the `CalculateCalibration` method.  Instead, I would recommend renaming the `CalculateCalibration` to `OptimizeCalibration` and creating a new method called `CalculateCalibration` that takes as input the `components_to_concentrations` and other structures for tracking the "outlier" and user "excluded" points, the quantitation method parameters, and simply calls the OpenMS methods for fitting the curve without running through the optimization process.

src/smartpeak/source/core/SequenceSegmentProcessors/CalculateCalibration.cpp
```diff
@@ -58,33 +71,82 @@ namespace SmartPeak
...
-
     // add in the method parameters
     OpenMS::AbsoluteQuantitation absoluteQuantitation;
     Utilities::setUserParameters(absoluteQuantitation, params_I);
 
     absoluteQuantitation.setQuantMethods(sequenceSegmentHandler_IO.getQuantitationMethods());
     std::map<std::string, std::vector<OpenMS::AbsoluteQuantitationStandards::featureConcentration>> components_to_concentrations;
     std::map<std::string, std::vector<OpenMS::AbsoluteQuantitationStandards::featureConcentration>> outlier_components_to_concentrations;
-    for (const OpenMS::AbsoluteQuantitationMethod& row : sequenceSegmentHandler_IO.getQuantitationMethods()) {
+    std::map<std::string, std::vector<OpenMS::AbsoluteQuantitationStandards::featureConcentration>> excluded_components_to_concentrations;
```
dmccloskey, line 96: Rebuilding these structures at each call (along with their modifications below) will be quite resource intensive.  I would recommend creating a separate method in the `SequenceSegmentHandler` to build the `features_concentrations_pruned` and remove all 0 values as done on lines 89 to 120 in the original code which can be used by both `OptimizeCalibration` and `CalculateCalibration` as needed.

src/smartpeak/source/core/SessionHandler.cpp
```diff
@@ -2007,121 +2060,122 @@ namespace SmartPeak
...
-        result.feature_raw_data.clear();
-        result.series_names.clear();
-        for (const auto& sequence_segment : sequence_handler.getSequenceSegments()) {
-          // Extract out raw data used to make the calibrators found in `StandardsConcentrations`
-          std::map<std::string, std::pair<std::vector<float>, std::vector<std::string>>> stand_concs_map; // map of x_data and sample_name for a component
-          for (const auto& stand_concs : sequence_segment.getStandardsConcentrations())
+      // LOGD << "Making the calibrators data for plotting";
+      // Update the axis titles and clear the data
+      result.x_axis_title = "Concentration ratio";
+      result.y_axis_title = sequence_handler.getSequenceSegments().at(0).getQuantitationMethods().at(0).getFeatureName() + " response";
```
dmccloskey: I would recommend `ratio` instead of response.

src/smartpeak/source/core/SessionHandler.cpp
```diff
@@ -2007,121 +2060,122 @@ namespace SmartPeak
...
+            std::vector<float> y_fit_data;
+            if (quant_method.getNPoints() > 0)
+            {
+              for (const auto& ratio : stand_concs_map.at(quant_method.getComponentName()).first) {
+                // TODO: encapsulate in its own method e.g. sequenceSegmentProcessor
+                // TODO: check that the calibration actually found a best fit (and set to all 0 if not)
+                // calculate the absolute concentration
+                OpenMS::TransformationModel::DataPoints data;
+                OpenMS::TransformationDescription tmd(data);
+                tmd.fitModel(quant_method.getTransformationModel(), quant_method.getTransformationModelParams());
```
dmccloskey, line 2109: Note that this would be a part of the method I refer to above in the context of the new `CalculateCalibration` method calculates which calculates the line of best fit using the user-specified points.

src/widgets/source/ui/CalibratorsPlotWidget.cpp
```diff
@@ -22,94 +22,376 @@
...
   void CalibratorsPlotWidget::displayParameters()
   {
+    if (component_cstr_.empty())
+    {
+      return;
+    }
+
+    static const std::vector<std::string> output_parameters = 
+    {
+      "llod",
```
dmccloskey, line 209: Currently llod and ulod are not calculated...

src/widgets/source/ui/CalibratorsPlotWidget.cpp
```diff
@@ -22,94 +22,376 @@
...
 
 #include <SmartPeak/ui/CalibratorsPlotWidget.h>
+#include <SmartPeak/core/SequenceSegmentProcessors/CalculateCalibration.h>
+#include <SmartPeak/core/Utilities.h>
 #include <implot.h>
 
 namespace SmartPeak
 {
+
+  void CalibratorsPlotWidget::recomputeCalibration()
```
dmccloskey, line 60: Overall, I think the updates to this widget are quite nice particularly the pop up of the chromatogram when selecting a point.  A couple of comments:

1. The SequenceSegment should be specified in e.g., a dropdown near the top of the widget where the component is specified.  This is because each sequentSegment can have its own quantitation method.

2. I would recommend two different user options for refitting: 1) OptimizeCalibration where all of the quantitation methods are recalculated using the SequenceSegment method `OptimizeCalibration`; and 2) ReCalculateCalibration where only the individual quantitation method is recalculated based on the previously optimized points and the points manually specified by the user to be included or excluded using the SequenceSegment method `CalculateCalibration`.

3. It might make sense to add a visual cue that alerts the user to the fact the the parameters are for the `OptimizeCalibration` method and the quantitation_method parameters are for both the `OptimizeCalibration` and `CalculateCalibration` methods.

Note that depending upon the user scenario, it might make sense to only allow for `CalculateCalibration`.  In that case, the parameters (i.e., for `OptimizeCalibration`) can be safely ignored and would not need to be shown to the user.

dmccloskey (changes requested):
I left some very detailed comments based on our previous conversation.  Please let me know if you would like to discuss more.

src/smartpeak/source/core/SessionHandler.cpp
bertrandboudaud: ok, done.

src/widgets/source/ui/CalibratorsPlotWidget.cpp
bertrandboudaud: so there is no possibility for the user to exclude points for the OptimizeCalibration?

dmccloskey: It should be possible for the user to exclude points for the OptimizeCalibration method.  Depending upon how the structure that links the standard with quantitation is built (see comment about making a separate method for that), an additional check to exclude user defined points should be possible.

This could cause problems down the road though if the user wanted to include all points for a fresh optimization.  So perhaps an option to add all points back in would be needed.

bertrandboudaud: Changed applied

src/smartpeak/source/core/SequenceSegmentProcessors/StoreQuantitationMethods.cpp
```diff
@@ -60,9 +60,30 @@ namespace SmartPeak
...
   };
 
+  bool StoreQuantitationMethods::onFilePicked(const std::filesystem::path& filename, ApplicationHandler* application_handler)
+  {
+    if (application_handler->sequenceHandler_.getSequenceSegments().size() == 0)
+    {
+      LOGE << "File cannot be written without first loading the sequence.";
+      return false;
+    }
+    auto& sequence_segment = application_handler->sequenceHandler_.getSequenceSegments().at(0);
```
dmccloskey, line 79: How is one able to store quantitationMethods from different sequenceSegments?

dmccloskey (commented):
The changes look nice.

I had a quick question.

Also, what was the section that you thought was "messy"?

src/smartpeak/source/core/SequenceSegmentProcessors/StoreQuantitationMethods.cpp
bertrandboudaud: Yes, actually it cannot support multiple segments at the moment. indeed. I would need to think about it, how to make it possible to the user to select the sequence segment (which could be tedious if you want to export all segments and if you have a lot) or maybe he will have to select a folder instead where we would export all the segments (probably the most easy to do).

bertrandboudaud:
> Also, what was the section that you thought was "messy"?

It's removeExcludedFromMatchingPoints but actually, I just removed it now, it's not used anymore with the latest changes

dmccloskey (approved):

src/smartpeak/source/core/SequenceSegmentProcessors/StoreQuantitationMethods.cpp
dmccloskey: I will make a ticket for this.

PR #520, opened 2022-03-11
Feat/processor inputs outputs

### LINKED ISSUE #493: Improved workflow step display
We want to improve the workflow step display.

The target goal is to develop a whole graph, but as a first step we can, keep that flat organization and improve it to prepare it:

1. Be able to distinguish raw_data_processor from sample_group_processors: different color, larger boxes with a header to easily make the difference.
2. Add input and output colored markers for what is produced and consumed.

bertrandboudaud:
Added inputs and output items to the workflow steps.

Also added Current Session Node on top (at the moment, it shows all possible outputs, even if the data there is empty: could be a next step to display it only when data is there)

Improved a little bit the dragging node handling.

InputOutputs

### DIFF
--- a/src/smartpeak/source/core/ApplicationHandler.cpp
+++ b/src/smartpeak/source/core/ApplicationHandler.cpp
@@ -174,6 +174,18 @@ namespace SmartPeak
     return (description ? description->getDescription() : "");
   }
 
+  std::set<std::string> ApplicationHandler::Command::getInputs() const
+  {
+    const auto description = getIProcessorDescription();
+    return (description ? description->getInputs() : std::set<std::string>());
+  }
+
+  std::set<std::string> ApplicationHandler::Command::getOutputs() const
+  {
+    const auto description = getIProcessorDescription();
+    return (description ? description->getOutputs() : std::set<std::string>());
+  }
+
   ApplicationHandler::Command::CommandType ApplicationHandler::Command::getType() const
   {
     return type;
--- a/src/smartpeak/source/core/RawDataProcessors/FilterFeatures.cpp
+++ b/src/smartpeak/source/core/RawDataProcessors/FilterFeatures.cpp
@@ -35,6 +35,16 @@
 namespace SmartPeak
 {
 
+  std::set<std::string> FilterFeatures::getInputs() const
+  {
+    return { "Feature Filter", "Targeted Experiment" };
+  }
+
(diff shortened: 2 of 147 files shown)


src/smartpeak/source/core/SequenceSegmentProcessors/CalculateCalibration.cpp
```diff
@@ -37,6 +37,16 @@
 
 namespace SmartPeak
 {
+  std::set<std::string> CalculateCalibration::getInputs() const
+  {
+    return { "Features"};
```
dmccloskey: "Standards Concentrations" and "Quantitation Methods" should also be inputs I believe.

src/smartpeak/source/core/ApplicationHandler.cpp
```diff
@@ -174,6 +174,18 @@ namespace SmartPeak
...
+  std::set<std::string> ApplicationHandler::Command::getInputs() const
+  {
+    const auto description = getIProcessorDescription();
+    return (description ? description->getInputs() : std::set<std::string>());
+  }
+
+  std::set<std::string> ApplicationHandler::Command::getOutputs() const
+  {
+    const auto description = getIProcessorDescription();
+    return (description ? description->getOutputs() : std::set<std::string>());
```
dmccloskey, line 186: Nice one liner.

src/smartpeak/source/core/RawDataProcessors/FilterFeatures.cpp
```diff
@@ -35,6 +35,16 @@
 namespace SmartPeak
 {
 
+  std::set<std::string> FilterFeatures::getInputs() const
+  {
+    return { "Feature Filters", "Targeted Experiment" };
```
dmccloskey: "Features" is missing I think

dmccloskey (changes requested):
The updates to the workflow widget look very nice.  I found a couple of missing inputs in the `getInputs` methods that I noted inline.

src/smartpeak/source/core/SequenceSegmentProcessors/CalculateCalibration.cpp
bertrandboudaud: indeed. thanks

dmccloskey (approved):

PR #524, opened 2022-03-18
Fix/GitHub 522 export quantmethods

### LINKED ISSUE #522: Export of QuantitationMethods from difference SequenceSegments
Yes, actually it cannot support multiple segments at the moment. indeed. I would need to think about it, how to make it possible to the user to select the sequence segment (which could be tedious if you want to export all segments and if you have a lot) or maybe he will have to select a folder instead where we would export all the segments (probably the most easy to do).

_Originally posted by @bertrandboudaud in https://github.com/AutoFlowResearch/SmartPeak/pull/515#discussion_r829480918_

bertrandboudaud:
Fixed Exportation of quantitation methods.

The user have to choose one directory, each file, for each sequence segment is created inside.

Note: in this PR, the examples are also fixed, they now use OptimizeCalibration instead of FitCalibration.


dmccloskey (approved):
Looks good so far.

PR #527, opened 2022-03-22
Fix/github521 create output folder

### LINKED ISSUE #521: Workflow fails if feature (output) directory does not exists
**Describe the bug**
If output directory does not exist, an error is displayed and workflow fails. This is ok, but the error is not 100% clear (failed to write feature file)

Would be nice to either:
- prevent the user to run workflow (but this could limit this feature to only GUI, not CLI).
- improve error message.
- or automatically create the output directory if it does not exists.

bertrandboudaud:
Currently, when the input or output directories are not there the error is not easy to understand. This fixes it, and also try to create the output directory if it does not exists.


dmccloskey (approved):
Nice additions. And in particular the unit test.

PR #528, opened 2022-03-23
Improving log messages

bertrandboudaud:
these modifications aim to both get the log message more understandable and also to clear the code, the creation of a processor, by removing redundant code.

each processor class now implements doProcess() instead of process and the parent processor class will log the message, take care of a possible error message and call the doProcessr() method.


dmccloskey (approved):
Nice changes.  It appears to clean things up quite a bit.

PR #530, opened 2022-03-24
Feat/rework n threads

bertrandboudaud:
Number of threads is now a parameter for the workflow manager, and not a user parameter anymore.

We now display the number of threads in the run workflow dialog box

image

CLI has been also impacted.

about CLI, tests revealed a bug when trying to override the workflow. this is fixed in this PR.


dmccloskey (approved):
Good catch with the updates to the CLI.

PR #531, opened 2022-03-24
Fix inclusion/exclusion of points for multiple Calibration display

### LINKED ISSUE #525: Fit calibration Cannot exclude point correctly when 2 calibration curves are displayed
**Describe the bug**
Cannot exclude point correctly when 2 calibration curves are displayed

**To Reproduce**
1. Execute "HPLC_UV_Standards"
2. Got to Veix Calibration
3. Select 2 Ttransistions in TRansitions Explorer, that would show 2 the calibration curves
4. Select a point from the second calibration, and try to exclude it.
5. The point not excluded.

**Expected behavior**
the point would be excluded.

**Version information**
1.21

**Additional context**
Note that if you select the right component on the combo box, you can indeed exclude the point correctly. But in that case, another incorrect behavior is that the component will be reset to the 1st one once the point is excluded (it may be considered as a different bug)

bertrandboudaud:


src/widgets/source/ui/CalibratorsPlotWidget.cpp
```diff
@@ -725,8 +724,9 @@ namespace SmartPeak
                   user_excluded_points_.erase(std::remove(user_excluded_points_.begin(), user_excluded_points_.end(), sample_and_serie), user_excluded_points_.end());
                   if (selected_action_ == EActionFitCalibration)
                   {
-                    recomputeCalibration();
-                  }                }
+                    recomputeFitCalibration(serie_name);
```
dmccloskey, line 727: It looks like I did not catch this last time but I think it should be "series" and not "serie" for the variable names?

src/widgets/source/ui/CalibratorsPlotWidget.cpp
```diff
@@ -57,39 +57,37 @@ namespace SmartPeak
     }
   }
 
-  void CalibratorsPlotWidget::recomputeCalibration()
+  void CalibratorsPlotWidget::recomputeFitCalibration(const std::string& component_name)
```
dmccloskey, line 60: I believe the "sequence_segment_name" would be needed here as well?  Or is that not correct because I see an internal variable on line 71 below called "sequence_segment_name_" that is specified before the call to ``fit_calibration.process()`?

src/widgets/source/ui/CalibratorsPlotWidget.cpp
bertrandboudaud: it's about only one serie (singular). it looks ok for me unless maybe i didn't get your comment correctly

src/widgets/source/ui/CalibratorsPlotWidget.cpp
bertrandboudaud: Only the component name is actually needed. it may not be easy to see in the code there.

when multiple curves are displayed on the plot. we used the component selected on the left panel. but it has to be the component of the point selected by the user.
However, about the sequence segment, all the curves from the plot belong to the same sequence segment. there is no ambiguity. so we can stay using the sequence segment selected on the left panel.
i think it's ok, but we can say it's a bit disturbing, i can add that parameter, but i will just pass the value of the left panel there.

src/widgets/source/ui/CalibratorsPlotWidget.cpp
dmccloskey: "series" is both singular and plural.  Anyway, very minor.  We can fix it in a subsequent PR.

dmccloskey (approved):

PR #532, opened 2022-04-05
Fixing Exportation for sequence segment data

bertrandboudaud:
This PR actually fixes 3 things:

- Filter were loaded only for the 1st sequence segment. now it loads for all sequence segments.
- Filter were exported only for the 1st sequence segment. now it exports for all sequence segments.
- Fix file directory file dialog on Mac OS when exporting.

This still does not fix the STORE_FEATURE_BACKGROUND_ESTIMATION processor.

### DIFF
--- a/src/examples/source/SmartPeakGUI.cpp
+++ b/src/examples/source/SmartPeakGUI.cpp
@@ -523,22 +523,22 @@ int main(int argc, char** argv)
             {std::make_shared<StoreValidationData>(), true},
             // {std::make_shared<StoreStandardsConcentrations>(), true}, // not implemented
             {std::make_shared<StoreQuantitationMethods>(true), false},
-            {std::make_shared<StoreFeatureFilters>(FeatureFiltersUtilsMode::EFeatureFiltersModeComponent, true), true},
-            {std::make_shared<StoreFeatureFilters>(FeatureFiltersUtilsMode::EFeatureFiltersModeGroup, true), true},
-            {std::make_shared<StoreFeatureQCs>(FeatureFiltersUtilsMode::EFeatureFiltersModeComponent, true), true},
-            {std::make_shared<StoreFeatureQCs>(FeatureFiltersUtilsMode::EFeatureFiltersModeGroup, true), true},
-            {std::make_shared<StoreFeatureRSDFilters>(FeatureFiltersUtilsMode::EFeatureFiltersModeComponent, true), true},
-            {std::make_shared<StoreFeatureRSDFilters>(FeatureFiltersUtilsMode::EFeatureFiltersModeGroup, true), true},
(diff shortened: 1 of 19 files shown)


src/examples/source/SmartPeakGUI.cpp
```diff
@@ -523,22 +523,22 @@ int main(int argc, char** argv)
...
-            {std::make_shared<StoreFeatureBackgroundFilters>(FeatureFiltersUtilsMode::EFeatureFiltersModeComponent, true), true},
-            {std::make_shared<StoreFeatureBackgroundFilters>(FeatureFiltersUtilsMode::EFeatureFiltersModeGroup, true), true},
-            {std::make_shared<StoreFeatureBackgroundQCs>(FeatureFiltersUtilsMode::EFeatureFiltersModeComponent, true), true},
-            {std::make_shared<StoreFeatureBackgroundQCs>(FeatureFiltersUtilsMode::EFeatureFiltersModeGroup, true), true},
-            {std::make_shared<StoreFeatureRSDEstimations>(FeatureFiltersUtilsMode::EFeatureFiltersModeComponent, true), true},
-            {std::make_shared<StoreFeatureRSDEstimations>(FeatureFiltersUtilsMode::EFeatureFiltersModeGroup, true), true},
-            {std::make_shared<StoreFeatureBackgroundEstimations>(FeatureFiltersUtilsMode::EFeatureFiltersModeComponent, true), true},
-            {std::make_shared<StoreFeatureBackgroundEstimations>(FeatureFiltersUtilsMode::EFeatureFiltersModeGroup, true), true}
+            {std::make_shared<StoreFeatureFilters>(FeatureFiltersUtilsMode::EFeatureFiltersModeComponent, true), false},
```
dmccloskey: FeatureFilters/QCs (Lines 526 to 529) have only one copy per sequence.

dmccloskey (changes requested):

src/examples/source/SmartPeakGUI.cpp
bertrandboudaud: Thank you. I don't see any code that assume that (including old revision). but I assume it's implicit information that does not require any special treatment. I had made the change so that when you export these data, you will export only one file.

dmccloskey (commented):
Sorry I missed this in my previous comment.  The Feature[RSD/Background][Filters/QCs]Component[Groups] should also be a single file per sequence.  These would be lines 530 to 537 in SmartPeakGUI.cpp

bertrandboudaud:
> Sorry I missed this in my previous comment. The Feature[RSD/Background][Filters/QCs]Component[Groups] should also be a single file per sequence. These would be lines 530 to 537 in SmartPeakGUI.cpp

ok, fixed. thank you

dmccloskey (approved):
Looks good.

PR #536, opened 2022-05-06
Fix/store estimations

bertrandboudaud:
Fixed Store STORE_FEATURE_BACKGROUND_ESTIMATIONS (and other similar processors)

Improved storing processors code with this behavior:
- Store* processor can:
	- Store externally ("non embedded")
	- Store internally (in the session, or "embbeded")
- When we save the Session we want to set them embedded, or not, depending on what we set in the session window.
- When we use a Store* processor in a workflow, we want to save externally only
- When we use a Store* processor from the export menu of Smartpeak we want to save externally only


dmccloskey (approved):
👍

PR #540, opened 2022-05-16
Removed scan explorer window

bertrandboudaud:
The scan explorer window actually introduces inconsistency. The user can select scan to plot, but there is no way to determine which sample this scan belongs to. To fix that we would need to add multiple controls that would complicate the user path.
in the end, in 99% of the cases, the user just wants to plot the whole scan. so we just remove that window and consider all scans being selected by the user and to be plotted.


dmccloskey (approved):
👍

PR #541, opened 2022-05-18
Fix/missing report metadata values

bertrandboudaud:
When creating reports, some metadata values are empty like chemical_formula, description, modification ...
we only considered the float conversion of metadata values which in some case contain strings.

### DIFF
--- a/src/smartpeak/source/core/Utilities.cpp
+++ b/src/smartpeak/source/core/Utilities.cpp
@@ -251,6 +251,57 @@ namespace SmartPeak
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/smartpeak/source/core/Utilities.cpp
```diff
@@ -251,6 +251,53 @@ namespace SmartPeak
     }
   }
 
+  CastValue Utilities::OpenMSDataValueToCastValue(
```
dmccloskey, line 254: Nice work on this.

dmccloskey (approved):
Nice additions with OpenMS value casting, and better handling of the float data.

PR #544, opened 2022-06-21
LoadMSP

dmccloskey:
# Description
- Add a new RawDataProcessor called LoadMSP that loads MSP files into a `library` member of RawDataHandler that is shared across all Injections
- Modifed MatchSpectra to call the `library` member instead of loading separate files from disk
- Added LoadMSP to the sequence initiation method so that users can load the library along with the rest of their sequence


bertrandboudaud (approved):
This is nice :)

My only concern is about the word "Library" used in this PR. could it be more precise (specifying what it contains)? This is a generic word. And perhaps later we would introduce other types of libraries.

src/smartpeak/source/core/RawDataProcessors/MatchSpectra.cpp
```diff
@@ -38,7 +38,7 @@ namespace SmartPeak
 
   std::vector<std::string> MatchSpectra::getRequirements() const
   {
-    return { "sequence", "traML" };
+    return { "sequence", "traML", "msp"};
   }
 
   std::set<std::string> MatchSpectra::getInputs() const
```
bertrandboudaud, line 44: Maybe MatchSpectra should have "Library" as an input

bertrandboudaud (changes requested):

src/smartpeak/source/core/RawDataProcessors/MatchSpectra.cpp
dmccloskey: I wanted to talk to you about this.  I am not sure how the naming for requirements vs. inputs/outputs is set-up but it seems it needs to be called "msp" for the UI to work correctly.  I tried "Library" (which I agree should be renamed) previously and I was not able to launch workflows.

dmccloskey:
@bertrandboudaud In `LoadMSP.cpp` on Line 61 I have a line commented out dealing with the Observable structure because it would not build.  However, it appears in other RawDataHandlers using the same syntax.  It would be great to get your thoughts on that.

src/smartpeak/source/core/RawDataProcessors/MatchSpectra.cpp
bertrandboudaud: my comment is about the getInputs method. basically, at the moment the getInputs/getOutputs is just a matter of being able to draw the flow arrows in the workflow display. perhaps later it could have more impact.

about getRequirements it's a list of things (: filenames) that the user has to specify before launching the workflow with this processor. if the associated filename is not provided an error would be displayed before launching it.

bertrandboudaud: getRequirements could be renamed "getFilenamesRequirement"

dmccloskey: > my comment is about the getInputs method. basically, at the moment the getInputs/getOutputs is just a matter of being able to draw the flow arrows in the workflow display. perhaps later it could have more impact.
> 
> about getRequirements it's a list of things (: filenames) that the user has to specify before launching the workflow with this processor. if the associated filename is not provided an error would be displayed before launching it.

I see.  Good catch.

dmccloskey: > getRequirements could be renamed "getFilenamesRequirement"

Agree :)

bertrandboudaud:
> @bertrandboudaud In `LoadMSP.cpp` on Line 61 I have a line commented out dealing with the Observable structure because it would not build. However, it appears in other RawDataHandlers using the same syntax. It would be great to get your thoughts on that.

make SequenceHandler inherits from LibraryObservable.

https://github.com/AutoFlowResearch/SmartPeak/commit/a65f8e90650d11a5092a4b4aad7729bef147a9e9

Also you want to use it along with the UI, you'll have to add some code in the EventDispatcher so that your event will be forwarded to the UI thread (it's quite easy, just follow how it has been done for the other observers).
We can have a short talk if you decide to have UI to respond to this observer

dmccloskey:
> > @bertrandboudaud In `LoadMSP.cpp` on Line 61 I have a line commented out dealing with the Observable structure because it would not build. However, it appears in other RawDataHandlers using the same syntax. It would be great to get your thoughts on that.
> 
> make SequenceHandler inherits from LibraryObservable.
> 
> [a65f8e9](https://github.com/AutoFlowResearch/SmartPeak/commit/a65f8e90650d11a5092a4b4aad7729bef147a9e9)
> 
> Also you want to use it along with the UI, you'll have to add some code in the EventDispatcher so that your event will be forwarded to the UI thread (it's quite easy, just follow how it has been done for the other observers). We can have a short talk if you decide to have UI to respond to this observer

Let's discuss this more next week.  It would be great to add a visualization of the library spectra to the `Spectra` widget view.

dmccloskey:
@bertrandboudaud It looks like there is an issue with code signing on Windows.  Has this error popped up before?

bertrandboudaud:
> @bertrandboudaud It looks like there is an issue with code signing on Windows. Has this error popped up before?

No I d'ont think I saw this one. I just started one again, let see if it's a temporary issue with CircleCI

PR #546, opened 2022-07-15
removed useless step in CI

bertrandboudaud:


dmccloskey (approved):
That is odd there was a duplication of steps.  Was it due to an error during a merge?

PR #551, opened 2022-12-29
Fix: Docs

dmccloskey:
# Features
- Added new tutorials for LCMS SRM and LCMS DDA
- Added new sections in Features and FAQs better describing the application features
- Added a placeholder section for the Server usage

# Fixes
- Types, grammar, etc. throughout the ReadTheDocs
- Combined several misc sections in the About SmartPeak part into the FAQs and Features
- Updated the SmartPeak GUI usage

# Todos
- installation instructions for MacOS
- updated developer documentation for e.g., make releases section
- new developer documentation for widgets and parameters
- updated screenshots using the latest verion of SmartPeak


bertrandboudaud:
That's a great improvement in the documentation!😀
Would be nice if we can find some tips to share some data between source code and some part of the documentation. Like for example generate the list of processors from the c++ code. Or the other way using doc in the workflow presets description and profit from the doc format (through a viewer). But this is another story.
I think it can be merged openms fails to compile now but it's not related to the doc.

bertrandboudaud (approved):

dmccloskey:
> Would be nice if we can find some tips to share some data between source code and some part of the documentation. Like for example generate the list of processors from the c++ code. Or the other way using doc in the workflow presets description and profit from the doc format (through a viewer).

Let's talk more about your idea next time we meet.  It sounds good to me.

PR #556, opened 2023-01-28
Set focus on the first declared windows in the workflow preset

bertrandboudaud:
For example, for HPLC_UV_Unkowns,
statistics and transitions will be focused

image

### DIFF
--- a/src/widgets/source/ui/SplitWindow.cpp
+++ b/src/widgets/source/ui/SplitWindow.cpp
@@ -33,7 +33,7 @@ namespace SmartPeak
     if (reset_layout_)
     {
       current_layout_.clear();
-      current_layout_ = new_layout;
+      current_layout_ = new_layout_;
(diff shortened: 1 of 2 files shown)


src/widgets/source/ui/SplitWindow.cpp
```diff
@@ -70,6 +70,18 @@ namespace SmartPeak
     }
   }
 
+  void SplitWindow::focusFirstWindow(const std::vector<std::tuple<std::shared_ptr<Widget>, bool>>& windows)
+  {
+    for (auto& [widget, visible] : windows)
+    {
+      if (widget->visible_)
+      {
+        ImGui::SetWindowFocus(widget->title_.c_str());
```
dmccloskey, line 79: Cool!

dmccloskey (approved):
Nice update!

PR #558, opened 2023-02-19
Fix for FitFeaturesEMG

dmccloskey:
# Fixes
- swapped `std::cout` for `LOGD`
- changed how `FitFeaturesEMG` updates the ConvexHull to accommodate a bug in OpenMS Feature::getConvexHulls() which returns the bounding box (4 points that outline the feature) and not the actual hull points when more than a single convex hull is detected, which does not allow for plotting.


dmccloskey:
@bertrandboudaud Do you have any idea what the build errors are?

bertrandboudaud:
> @bertrandboudaud Do you have any idea what the build errors are?

on linux we can see this error:
/root/SmartPeak/src/smartpeak/source/core/RawDataProcessors/FitFeaturesEMG.cpp:156:87:   required from here
/root/SmartPeak/superbuild/Dependencies/Source/plog/include/plog/Record.h:162:23: error: ambiguous overload for 'operator<<' (operand types are 'plog::util::nostringstream' {aka 'std::__cxx11::basic_ostringstream<char>'} and 'const OpenMS::DataValue')
  162 |             m_message << data;
      |             ~~~~~~~~~~^~~~~~~

src/smartpeak/source/core/RawDataProcessors/FitFeaturesEMG.cpp
```diff
@@ -130,29 +130,33 @@ namespace SmartPeak
...
         if (peak_integral < 0) { peak_integral = 0; }
         if (peak_apex_int < 0) { peak_apex_int = 0; }
 
-        std::cout << "Intensity: " << subfeature.getIntensity() << "\t" << peak_integral << "\n";
-        std::cout << "peak_apex_position: " << subfeature.getMetaValue("peak_apex_position") << "\t" << pa.apex_pos << "\n";
-        std::cout << "peak_apex_int: " << subfeature.getMetaValue("peak_apex_int") << "\t" << peak_apex_int << "\n";
-        std::cout << "area_background_level: " << subfeature.getMetaValue("area_background_level") << "\t" << pb.area << "\n";
-        std::cout << "noise_background_level: " << subfeature.getMetaValue("noise_background_level") << "\t" << pb.height << "\n\n";
+        LOGD << "Intensity: " << subfeature.getIntensity() << "\t" << peak_integral;
+        LOGD << "peak_apex_position: " << subfeature.getMetaValue("peak_apex_position") << "\t" << pa.apex_pos;
```
bertrandboudaud: Looks like there is an ambiguity at least on linux for the 4 following lines. Perhaps you can try to cast to either int, long int or bool

dmccloskey: I see.  OK I should have looked a bit deeper into the logs.  I will take care of this in a few hours I think on the next push.

dmccloskey: @bertrandboudaud I fixed the build issues on MacOS and Linux.  However, Windows is failing, but the build output is more than 50 MB so I am not sure what the issue is.  Builds on my machine :).  Have you used the `store_artifacts` to save the build log or is there a simpler way to find out what the issue is?

dmccloskey: @bertrandboudaud I figured out what the issue was.

bertrandboudaud (approved):

bertrandboudaud (approved):
Nice :)

